PR #101, opened 2023-12-11
Update issue templates

jkaufman-mitre:


jkaufman-mitre:
@buidav Realized there was a template missing from the other PR.

tmcomeau (approved):

LaurenBassett (approved):

jkaufman-mitre:
What templates do you want as there does not seem to be an option to have no templates. I at least need a way to be able to create a blank issue. @buidav As of now, I am not able to get done what I need to get done.

tmcomeau:
Thanks for the chat @buidav. @jkaufman-mitre sounds like if we can re-introduce the 2 existing templates exactly as they were setup previously, plus your new one (try to stick with the established conventions for that too to the extend possible to maintain consistency) we should be gtg!

Issue #127, opened 2024-01-03
Wording Changes in Drive_Docs


jkaufman-mitre:
The following changes are needed based on an RFC comment:

- DRIVEDOCS.1.2v0.1: Shorten the policy statement to "Agencies SHOULD disable users' receiving files from outside of the
organization's domain."
- DRIVEDOCS.2.5v0.1: Pending determination of what changes are needed to DRIVEDOCS.2.5v0.1
- DRIVEDOCS.4.1v0.1: Remove the following wording from the policy statement: "to restrict information sharing and prevent data leakage."
- DRIVEDOCS.5.1v0.1: Remove the following wording from the policy statement: "with the exception of those that are approved within the organization."
- Policy Group 6: Re-word the intro to the following: "This section covers the implications of not disabling Drive for Desktop and considerations around authorized device use."

PR #128, opened 2024-01-03
Changes addressing in issue 127

jkaufman-mitre:
Addressed the first set of comments related to DRIVE_DOCS as outlined in Issue 127. This issue covers all of the changes in that issue.

### DIFF
--- a/baselines/Google Drive and Docs Minimum Viable Secure Configuration Baseline v0.1.md
+++ b/baselines/Google Drive and Docs Minimum Viable Secure Configuration Baseline v0.1.md
@@ -56,7 +56,7 @@ Agencies SHOULD disable sharing outside of the organization's domain.
   - [T1537: Transfer Data to Cloud Account](https://attack.mitre.org/techniques/T1537/)
 
 #### GWS.DRIVEDOCS.1.2v0.1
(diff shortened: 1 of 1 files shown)


jkaufman-mitre:
Fixes #127

LaurenBassett (approved):

tmcomeau (approved):

baselines/Google Drive and Docs Minimum Viable Secure Configuration Baseline v0.1.md
```diff
@@ -393,7 +393,7 @@ To configure the settings for add-ons:
 
 ## 6. Drive for Desktop
 
-This section covers that Google Drive for Desktop, if not disabled entirely, should only be allowed on authorized devices.
+This section covers the implications of not disabling Drive for Desktop and considerations around authorized device use.
```
adhilto, line 396: This wording feels a little awkward. Also, it might be more useful to use this space to describe what Drive for Desktop is, as not all users will be familiar with it.

How about something like this?

```suggestion
This section addresses Drive for Desktop, a feature that enables users to interact with their Drive files directly through their desktop's file system, rather than through the browser.
```

adhilto (changes requested):
See my above comment.

adhilto (approved):

jkaufman-mitre:
Fixed the last modified dates. @buidav

PR #209, opened 2024-02-21
Resolve Drive & Docs Baseline dependency Policy Issues

### LINKED ISSUE #70: Drive/Docs Baseline Dependent Baseline Policy Issues
There's been a few issues about this, but I wanted to put everything in one place. 

The baseline policies for GWS.DRIVEDOCS.1 are not contradictory, but not all are applied in every situation. 

GWS.DRIVEDOCS.1.1v1 - Agencies SHOULD disable sharing outside of the organization‚Äôs domain 
Determines which of the following baselines should be 'active' and being checked. In my opinion, we shouldn't have this as a pass/fail, as we recommend that it is turned off, but have baselines established when it is turned on. 

**if sharing outside the organization is disabled (GWS.DRIVEDOCS.1.1v1 is 'compliant')**
Then, GWS.DRIVEDOCS.1.2v1 should be evaluated (If disabling sharing outside of the organization's domain, then agencies SHALL also disable users' receiving files from outside of the organization's domain.) 

Additionally, 
GWS.DRIVEDOCS.1.3.1v1 and GWS.DRIVEDOCS.1.4.1v1 do NOT need to be checked. 

**if sharing outside the organization is enabled (GWS.DRIVEDOCS.1.1v1 is 'non-compliant')**
Then, GWS.DRIVEDOCS.1.2.1v1 does not need to be checked 

and 
GWS.DRIVEDOCS.1.3.1v1 and GWS.DRIVEDOCS.1.4.1v1 need to be checked. 

Currently, all baselines are being evaluated every time, which is causing extra failures when they don't need to be. 

I worry about how this would be extended to multiple OUs, but I think this is the core of where the logic inconsistencies are coming from.

snarve:
## 🗣 Description ##

Closes #70 

### 💭 Motivation and context

 If 1.1 is 'OFF' - sharing is not allowed, then the other checks are not applicable
 So all the other policies through 1.2 to 1.4 should be checked conditionally only if sharing is 'ON' 

## 🧪 Testing

Discussed these updates with @adhilto and also performed a preliminary test to check different use cases 

## ✅ Pre-merge Checklist

## ✅ Post-merge Checklist

### DIFF
--- a/rego/Drive.rego
+++ b/rego/Drive.rego
@@ -17,7 +17,7 @@ NonCompliantOUs1_1 contains OU if {
     Events := utils.FilterEvents(LogEvents, "SHARING_OUTSIDE_DOMAIN", OU)
     count(Events) > 0
     LastEvent := utils.GetLastEvent(Events)
-    AcceptableValues := {"SHARING_NOT_ALLOWED", "INHERIT_FROM_PARENT"}
+    AcceptableValues := {"SHARING_NOT_ALLOWED", "INHERIT_FROM_PARENT", "SHARING_NOT_ALLOWED_BUT_MAY_RECEIVE_FILES"}
     not LastEvent.NewValue in AcceptableValues
 }
 
@@ -59,7 +59,7 @@ NonCompliantOUs1_2 contains OU if {
     Events := utils.FilterEvents(LogEvents, "SHARING_OUTSIDE_DOMAIN", OU)
     count(Events) > 0
     LastEvent := utils.GetLastEvent(Events)
-    contains("SHARING_NOT_ALLOWED INHERIT_FROM_PARENT", LastEvent.NewValue) == false  
+    contains("SHARING_NOT_ALLOWED INHERIT_FROM_PARENT", LastEvent.NewValue) == false
 }
 
 tests contains {
@@ -100,7 +100,7 @@ NonCompliantOUs1_3 contains OU if {
     Events := utils.FilterEvents(LogEvents, "SHARING_OUTSIDE_DOMAIN", OU)
     count(Events) > 0
     LastEvent := utils.GetLastEvent(Events)
-    AcceptableValues := {"SHARING_ALLOWED_WITH_WARNING", "SHARING_NOT_ALLOWED", "INHERIT_FROM_PARENT"}
+    AcceptableValues := {"SHARING_ALLOWED_WITH_WARNING", "SHARING_NOT_ALLOWED", "INHERIT_FROM_PARENT", "SHARING_NOT_ALLOWED_BUT_MAY_RECEIVE_FILES"}
     not LastEvent.NewValue in AcceptableValues
 }
 
@@ -162,7 +162,7 @@ NonCompliantOUs1_4 contains OU if {
 
     AcceptableValues_A := {"NOT_ALLOWED", "INHERIT_FROM_PARENT"}
     not LastEvent_A.NewValue in AcceptableValues_A
-    AcceptableValues_B := {"SHARING_ALLOWED_WITH_WARNING", "SHARING_NOT_ALLOWED", "INHERIT_FROM_PARENT"}
+    AcceptableValues_B := {"SHARING_NOT_ALLOWED", "INHERIT_FROM_PARENT"}
     not LastEvent_B.NewValue in AcceptableValues_B
 }
 
--- a/rego/Drive.rego
+++ b/rego/Drive.rego
@@ -99,7 +99,7 @@ NonCompliantOUs1_3 contains OU if {
     Events := utils.FilterEvents(LogEvents, "SHARING_OUTSIDE_DOMAIN", OU)
     count(Events) > 0
     LastEvent := utils.GetLastEvent(Events)
-    contains("SHARING_ALLOWED_WITH_WARNING INHERIT_FROM_PARENT SHARING_NOT_ALLOWED", LastEvent.NewValue) == false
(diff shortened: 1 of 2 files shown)


adhilto (changes requested):
At a high level, changes look good. However, some of the current unit tests are broken by these changes. Please update those tests and add a few new ones to cover the new logic added by this PR.

snarve:
All unit tests are fixed and also updated the rego code.
This is ready for review, however, I have the following recommendation:

For policy 1:

1.1 and 1.2 can be combined into a single policy as both 1.1 and 1.2 can be checked using a single setting and use the same value for the setting.

Similarly, 1.3 and 1.4 can be combined. 1.4 will confirm 1.3 passes since 1.4 uses the same setting as 1.3 and an additional setting. So essentially 1.4 is already checking for 1.3 and 1.4.

LaurenBassett (approved):

adhilto:
> All unit tests are fixed and also updated the rego code. This is ready for review, however, I have the following recommendation:
> 
> For policy 1:
> 
> 1.1 and 1.2 can be combined into a single policy as both 1.1 and 1.2 can be checked using a single setting and use the same value for the setting.
> 
> Similarly, 1.3 and 1.4 can be combined. 1.4 will confirm 1.3 passes since 1.4 uses the same setting as 1.3 and an additional setting. So essentially 1.4 is already checking for 1.3 and 1.4.

@snarve Could you open an issue to track those recommendations?

rego/Drive.rego
```diff
@@ -59,7 +59,7 @@ NonCompliantOUs1_2 contains OU if {
     Events := utils.FilterEvents(LogEvents, "SHARING_OUTSIDE_DOMAIN", OU)
     count(Events) > 0
     LastEvent := utils.GetLastEvent(Events)
-    contains("SHARING_NOT_ALLOWED INHERIT_FROM_PARENT", LastEvent.NewValue) == false
+    contains("SHARING_NOT_ALLOWED INHERIT_FROM_PARENT", LastEvent.NewValue) == false
```
adhilto, line 62: Recommend removing trailing whitespace
```suggestion
    contains("SHARING_NOT_ALLOWED INHERIT_FROM_PARENT", LastEvent.NewValue) == false
```

snarve:
> > All unit tests are fixed and also updated the rego code. This is ready for review, however, I have the following recommendation:
> > For policy 1:
> > 1.1 and 1.2 can be combined into a single policy as both 1.1 and 1.2 can be checked using a single setting and use the same value for the setting.
> > Similarly, 1.3 and 1.4 can be combined. 1.4 will confirm 1.3 passes since 1.4 uses the same setting as 1.3 and an additional setting. So essentially 1.4 is already checking for 1.3 and 1.4.
> 
> @snarve Could you open an issue to track those recommendations?

New issue filed: https://github.com/cisagov/ScubaGoggles/issues/219

rego/Drive.rego
```diff
@@ -99,7 +100,8 @@ NonCompliantOUs1_3 contains OU if {
     Events := utils.FilterEvents(LogEvents, "SHARING_OUTSIDE_DOMAIN", OU)
     count(Events) > 0
     LastEvent := utils.GetLastEvent(Events)
-    contains("SHARING_ALLOWED INHERIT_FROM_PARENT", LastEvent.NewValue) == true
+    AcceptableValues := {"SHARING_ALLOWED_WITH_WARNING", "SHARING_NOT_ALLOWED", "INHERIT_FROM_PARENT"}
```
adhilto, line 103: `SHARING_NOT_ALLOWED_BUT_MAY_RECEIVE_FILES` needs to be included in the list of acceptable values. Without it, the following configuration results in a failure for 1.3:
image

```suggestion
    AcceptableValues := {"SHARING_ALLOWED_WITH_WARNING", "SHARING_NOT_ALLOWED", "INHERIT_FROM_PARENT",
        "SHARING_NOT_ALLOWED_BUT_MAY_RECEIVE_FILES"}
```

rego/Drive.rego
```diff
@@ -134,12 +136,34 @@ if {
...
+    count(Events_A) > 0
+    LastEvent_A := utils.GetLastEvent(Events_A)
+
+    Events_B := utils.FilterEvents(LogEvents, "SHARING_OUTSIDE_DOMAIN", OU)
+    count(Events_B) > 0
+    LastEvent_B := utils.GetLastEvent(Events_B)
+
+    AcceptableValues_A := {"NOT_ALLOWED", "INHERIT_FROM_PARENT"}
+    not LastEvent_A.NewValue in AcceptableValues_A
+    AcceptableValues_B := {"SHARING_ALLOWED_WITH_WARNING", "SHARING_NOT_ALLOWED", "INHERIT_FROM_PARENT"}
```
adhilto, line 165: `SHARING_ALLOWED_WITH_WARNING` should not be in the list of AcceptableValues_B. Otherwise, the following configuration results in a pass for 1.4:
image

```suggestion
    AcceptableValues_B := {"SHARING_NOT_ALLOWED", "INHERIT_FROM_PARENT"}
```

adhilto (changes requested):

rego/Drive.rego
snarve: The checkbox should be unselected for 1.2 (Allow users to receive files from users or...)

image

image

image

snarve: The checkbox for 1.2 should be deselected (Allow users to receive files....)

image

image

image

snarve: Ok, made the updates.
Also, noticed that for 1.1 "SHARING_NOT_ALLOWED_BUT_MAY_RECEIVE_FILES" should be added, so that 1.1 passes even when and if 1.2 fails. This way now all of them 1.1 through 1.4 are independently working. Hope this covers all the use cases and the outputs are as expected.

adhilto: Good catch about 1.1

adhilto (approved):

PR #311, opened 2024-06-07
Drive_Docs Changes to 1.6 and 6.1

### LINKED ISSUE #142: Drive Docs 6.1 Confusing
The GWS.DRIVEDOCS.6.1v0.1 policy currently reads as: "Agencies SHOULD either disable Google Drive for Desktop or only allow Google Drive for Desktop on authorized devices."

The rationale reads as: "Saving directly to the cloud reduces the risk of potentially losing data, which is beneficial when dealing with sensitive information and when trying to retain certain files."

These two sections seem to conflict with each other. Enabling Google Drive for Desktop makes it _easier_ to have cloud-only file storage. So the policy seems to be _against_ cloud-only while the rationale is _pro_ cloud-only.

I'd recommend the following changes:
1. Change the policy wording to clarify the true purpose. My recommendation: "Google Drive SHOULD only be enabled for authorized devices." Reasoning: the simplified logic makes it easier to understand the true purpose of the policy. It's not saying that you can't use Drive for Desktop, only that users shouldn't be able to connect their agency Google Drive to something like their personal computer. If an agency wants to disable Drive for Desktop entirely, that still meets the requirements of the policy; there are just no approved computers in that case.
2. Change the rationale so that the true security concerns are highlighted. My recommendation: "Some users may attempt to use Drive for Desktop to connect unapproved devices (e.g., a personal computer), to the agency's Google Drive. Even if done without malicious intent, this represents a security risk as the agency has no ability audit or protect such computers."

### LINKED ISSUE #289: GWS DriveDocs 1.6 v0.1 Policy Changes
After 0.2 Release

Policy re-evaluation might be needed because setting is enabled always, just need to make sure it's not set to public.

Might need to re-adjust rationale as well.

### LINKED ISSUE #302: GWS.DRIVEDOCS.1.6v0.1 clarification needed
The current baseline reads as: "Agencies SHALL enable access checking for file sharing outside of Docs or Drive."

However, this setting isn't an enable/disabled setting. In fact, it can't be disabled, so additional clarification is needed. See screenshot below for the options presented:
image

It seems like the important bit is that access check not allow sharing to the public. As such, I'd recommend rewording the baseline statement to something along the lines of: "Agencies SHALL NOT allow Access Checker to share files with the public."

### LINKED ISSUE #313: Clarify GWS.DRIVEDOCS.3.1v0.2
This policy currently reads as: "Agencies SHALL enable security updates for Drive files." This wording make it sound like [GWS.GMAIL.5.4v0.2](https://github.com/cisagov/ScubaGoggles/blob/main/baselines/Gmail%20Minimum%20Viable%20Secure%20Configuration%20Baseline%20v0.2.md#gwsgmail54v02), "apply future recommend settings." The rationale supports this interpretation: "Google may add new security features over time. Allowing security updates helps ensure that your files are protected with the latest features Google makes available."

However, digging deeper, it turns out the setting in question is about applying _a specific security update_ that has already been published, not about applying new updates as they come. See [Link-sharing security update to Drive](https://apps.google.com/supportwidget/articlehome?article_url=https%3A%2F%2Fsupport.google.com%2Fa%2Fanswer%2F10685032%3Fvisit_id%3D638533824698144528-3160863719&assistant_event=welcome&assistant_id=mega-bot-shared-drive&product_context=10685032&product_name=UnuFlow&trigger_context=a) for more information.

I recommend the following:
- Clarify the policy wording. Maybe "Agencies SHALL enable the security update for Drive files."
- Clarify the rationale.
- Swap out the link that's currently in the resources section for the one I pasted above. The current one doesn't work ("Sorry, this page can't be found.")

jkaufman-mitre:
## 🗣 Description ##

The following changes were made to the Drive_Docs baselines:
- Adjusted Drive_Docs 1.6 to not allow agencies to set access checker to public
- Adjusted Drive_Docs 6.1 to say that agencies SHOULD set Drive for Desktop to authorized devices only
- Added the changes to DRIVEDOCS.3.1v0.2

### 💭 Motivation and context
Closes #142.
Closes #289.
Closes #302.
Closes #313.

## ✅ Pre-merge Checklist

## ✅ Post-merge Checklist


baselines/Google Drive and Docs Minimum Viable Secure Configuration Baseline v0.2.md
```diff
@@ -98,10 +98,11 @@ Agencies SHALL disable making files and published web content visible to anyone
   - [T1537: Transfer Data to Cloud Account](https://attack.mitre.org/techniques/T1537/)
 
 #### GWS.DRIVEDOCS.1.6v0.2
-Agencies SHALL enable access checking for file sharing outside of Docs or Drive.
+Agencies SHOULD set access checking to recipients only.
 
 - _Rationale:_ The Access Checker feature can be configured to allows users to grant access to the public if a recipient is missing access, creating the potential for data leakage. This control mitigates this by only allowing access to be granted to recipients or the suggested target audience.
-- _Last modified:_ July 10, 2023
+- _Last modified:_ June 7, 2024
+- _Note:_ Agencies SHALL NOT set access checking to Recipients only, suggested target audience, or public (no Google account required).
```
adhilto, line 105: We shouldn't have a normative requirement, especially a SHALL, in a note. Notes are for informational bits, like clarifications or exceptions. Furthermore, ScubaGoggles isn't capable of checking polices that are simultaneously SHALL and SHOULD. On M365, we've gone through the work of decoupling all such policies (just a few left, for example see: https://github.com/cisagov/ScubaGear/issues/524). We shouldn't start adding them here.

Furthermore, I don't think the "suggested target audience" option is necessarily a bad thing. At first glance it seems like a very hand-wavy thing, leaving you wondering how Google determines who is in your target audiences; if that were the case I would agree that we should recommend against it. But it turns out that target audiences are actually a very precise thing that you have to configure. See [About target audiences](https://support.google.com/a/answer/9934697?hl=en.) The bottom line: you decide who goes into a target audience.
And per [Add or remove target audience members](https://support.google.com/a/answer/9934621):
- Non-Google users can't be added to target audience
- External users can only be added to a target audience if Drive allows sharing outside your organization (i.e., GWS.DRIVEDOCS.1.1v0.2).
Those already built-in requirements for target audiences minimizes the risk they present.

Therefore, I'd recommend taking this note and making it the main policy for GWS.DRIVEDOCS.1.6v0.2. Though if we did care enough about target audiences to include a soft recommendation against them here, this SHALL NOT statement should be split out to into its own policy.

Finally, currently the wording for the SHALL NOT statement can be read as listing 3 different things that access checking shouldn't be set to, i.e., "don't set it to recipients only, also don't set it to suggested target audience, also don't set it to public." Wrapping "Recipients only, suggested target audience, or public (no Google account required)" in quotes would make it more clear that we're referring to a single option. But that introduces the problem of using the precise "buttonology" in the policy statement, something we try to avoid to make the policies less brittle, especially with a string that long. The real important bit is that they shouldn't set it to public. One potential way of wording it that fixes these issues (and as an added bonus clarifies what this setting actually does):
>Access Checking SHALL NOT suggest sharing to the public (no access required).

baselines/Google Drive and Docs Minimum Viable Secure Configuration Baseline v0.2.md
```diff
@@ -367,10 +368,10 @@ This section addresses Drive for Desktop, a feature that enables users to intera
 ### Policies
 
 #### GWS.DRIVEDOCS.6.1v0.2
-Agencies SHOULD either disable Google Drive for Desktop or only allow Google Drive for Desktop on authorized devices.
+Google Drive for Desktop SHOULD only be enabled for authorized devices.
```
adhilto, line 371: Maybe this one is just personal preference, but I think moving the "only" makes it read better.
```suggestion
Google Drive for Desktop SHOULD be enabled only for authorized devices.
```

adhilto (changes requested):

baselines/Google Drive and Docs Minimum Viable Secure Configuration Baseline v0.2.md
jkaufman-mitre: We will take a look at it. However, "Recipients only, suggested target audience, or public (no Google account required)" is not 3 things it is 1 option in Google Workspace Admin Console. If you look at the setting all of these are listed in 1 selection option. @adhilto Please make sure to look more closely before making a comment about something.

adhilto: > However, "Recipients only, suggested target audience, or public (no Google account required)" is not 3 things it is 1 option in Google Workspace Admin Console

@jkaufman-mitre That's my exact point. The way it's currently worded in the baseline makes it sound like it could be 3 things. My suggested rewording removes that ambiguity.

jkaufman-mitre:
Added the changes to DRIVEDOCS.3.1v0.2 into this PR. Fixes #313

adhilto:
> Added the changes to DRIVEDOCS.3.1v0.2 into this PR. Fixes #313

@jkaufman-mitre with regard to #313, please also:
- Clarify the rationale. The current rationale was written through the lens that this setting would apply future security updates as Google releases them.
- Swap out the link that's currently in the resources section for the one I pasted in #313, or another one as you see fit. The current one doesn't work ("Sorry, this page can't be found.")

baselines/Google Drive and Docs Minimum Viable Secure Configuration Baseline v0.2.md
```diff
@@ -98,10 +98,11 @@ Agencies SHALL disable making files and published web content visible to anyone
   - [T1537: Transfer Data to Cloud Account](https://attack.mitre.org/techniques/T1537/)
 
 #### GWS.DRIVEDOCS.1.6v0.2
-Agencies SHALL enable access checking for file sharing outside of Docs or Drive.
+Agencies SHOULD set access checking to recipients only.
```
adhilto, line 101: A little bit more work needs to be done here to bring the policy statement, the note, and the implementation instructions in-sync with each other.

Per the policy statement, there is only one acceptable option:
image

Per the note, there are two acceptable options:
image

Finally, per the implementation, there is just one acceptable option:
image

baselines/Google Drive and Docs Minimum Viable Secure Configuration Baseline v0.2.md
```diff
@@ -260,17 +261,17 @@ This section covers whether a security update issued by Google will be applied t
 ### Policies
 
 #### GWS.DRIVEDOCS.3.1v0.2
-Agencies SHALL enable security updates for Drive files.
+Agencies SHALL enable the security update for Drive files.
 
-- _Rationale:_ Google may add new security features over time. Allowing security updates helps ensure that your files are protected with the latest features Google makes available.
+- _Rationale:_ By not enabling the resource key security update it creates the potential for an unauthorized access to files. Enabling this security update mitigates the risk by ensuring access is controlled properly.
```
adhilto, line 266: The first sentence doesn't work grammatically and the second promises too much (access can still be improperly controlled even if this update is applied).

adhilto (changes requested):
Thanks for these changes. A few more comments:

baselines/Google Drive and Docs Minimum Viable Secure Configuration Baseline v0.2.md
jkaufman-mitre: our policy statement says it SHOULD be set to that option, but the note is telling them that another acceptable option is the other one and not to set it to the one that includes public. The way we have it is the way we intended it to be.

baselines/Google Drive and Docs Minimum Viable Secure Configuration Baseline v0.2.md
mdueltgen: By not enabling the update to the resource key security update a user could potentially gain unauthorized access to files. Enabling this security update decreases risk of unauthorized access and data spillage by controlling access to files in Google Drive. 

@bnewlin-MITRE @adhilto I didn't change much from the original but addressed grammatical issues and the over promising

mdueltgen:
I updated the rationale statement for 3.1. Requesting review

baselines/Google Drive and Docs Minimum Viable Secure Configuration Baseline v0.2.md
```diff
@@ -98,10 +98,10 @@ Agencies SHALL disable making files and published web content visible to anyone
   - [T1537: Transfer Data to Cloud Account](https://attack.mitre.org/techniques/T1537/)
 
 #### GWS.DRIVEDOCS.1.6v0.2
-Agencies SHALL enable access checking for file sharing outside of Docs or Drive.
+Agencies SHALL set access checking to recipients only.
 
-- _Rationale:_ The Access Checker feature can be configured to allows users to grant access to the public if a recipient is missing access, creating the potential for data leakage. This control mitigates this by only allowing access to be granted to recipients or the suggested target audience.
-- _Last modified:_ July 10, 2023
+- _Rationale:_ The Access Checker feature can be configured to allow users to grant open access if a recipient is missing access, creating the potential for data leakage. This control mitigates this by only allowing access to be granted to recipients or the suggested target audience.
```
adhilto, line 103: Looks like we forgot to remove the "or suggested target audience" bit in the rationale.
```suggestion
- _Rationale:_ The Access Checker feature can be configured to allow users to grant open access if a recipient is missing access, creating the potential for data leakage. This control mitigates this by only allowing access to be granted to recipients.
```

drift-rules/GWS Drift Monitoring Rules - Drive and Docs.csv
```diff
@@ -4,16 +4,15 @@ GWS.DRIVEDOCS.1.2v0.2,"If disabling sharing outside of the organization's domain
...
 GWS.DRIVEDOCS.5.1v0.2,Agencies SHALL disable Add-Ons with the exception of those that are approved within the organization.,Admin Log Event,Change Drive Setting,ENABLE_DOCS_ADD_ONS,false,rules/00gjdgxs4d794jn,JK 08-02-23 @ 15:14
-GWS.DRIVEDOCS.6.1v0.2(a),Agencies SHOULD either disable Google Drive for Desktop or only allow Google Drive for Desktop on authorized devices.,Admin Log Event,Change Application Setting,	DriveFsSettingsProto drive_fs_enabled,false,rules/00gjdgxs0yziufl,JK 10-19-23 @ 13:47
-GWS.DRIVEDOCS.6.1v0.2(b),Agencies SHOULD either disable Google Drive for Desktop or only allow Google Drive for Desktop on authorized devices.,Admin Log Event,Change Application Setting,DriveFsSettingsProto company_owned_only_enabled,true,rules/00gjdgxs4ghyiin,JK 10-19-23 @ 14:01
+GWS.DRIVEDOCS.6.1v0.2,Google Drive for Desktop SHOULD only be enabled for authorized devices.,Admin Log Event,Change Application Setting,DriveFsSettingsProto company_owned_only_enabled,true,rules/00gjdgxs4ghyiin,JK 10-19-23 @ 14:01
```
adhilto, line 17: Just noticed that the current wording here doesn't match the wording used in the markdown
```suggestion
GWS.DRIVEDOCS.6.1v0.2,Google Drive for Desktop SHOULD be enabled only for authorized devices..,Admin Log Event,Change Application Setting,DriveFsSettingsProto company_owned_only_enabled,true,rules/00gjdgxs4ghyiin,JK 10-19-23 @ 14:01
```

adhilto (approved):
Baseline looks good and I've implemented the needed changes in the Rego. Just one minor comment about the drift rule, but I'm going to go ahead and approve.

mdueltgen (approved):

LaurenBassett (approved):

PR #322, opened 2024-06-21
Changed 5.1 to a SHOULD and added a note

### LINKED ISSUE #222: Pre-Requisite needs to be Added to Common Controls Policy Group 11
Based on the the google update below, a prerequisite needs to be added to Common Controls Policy Group 11 stating that the "Allow users to install Chat apps" needs to be set to "On".

[Google Blog Update Post](http://workspaceupdates.googleblog.com/2024/02/updates-to-google-workspace-marketplace.html)

jkaufman-mitre:
## 🗣 Description ##

Based on the google update that came out, changed Chat 5.1 to SHOULD to allow for an organization to allow users to install chat apps since Common Controls states to allow apps to be installed if it is allowlisted.

Fixes #222 

### 💭 Motivation and context

Google Update

## ✅ Pre-merge Checklist

## ✅ Post-merge Checklist

### DIFF
--- a/baselines/Google Chat Minimum Viable Secure Configuration Baseline v0.2.md
+++ b/baselines/Google Chat Minimum Viable Secure Configuration Baseline v0.2.md
@@ -206,10 +206,11 @@ This section covers preventing ordinary users from being able to install Chat ap
 ### Policies
 
 #### GWS.CHAT.5.1v0.2
-User-level ability to install Chat apps SHALL be disabled.
+User-level ability to install Chat apps SHOULD be enabled.
 
 - _Rationale:_ Unrestricted access to Chat apps can expose the agency to potential vulnerabilities present in those apps or apps developed with malicious intent. This policy prevents unauthorized app installations and promotes the protection of sensitive information.
 - _Last modified:_ July 10, 2023
+- _Note:_ Common Controls only allows apps that our allowlisted to be installed
 
 - MITRE ATT&CK TTP Mapping
   - [T1195:002: Supply Chain Compromise](https://attack.mitre.org/techniques/T1195/)


jkaufman-mitre:
@adhilto This PR is still waiting to be reviewed.

adhilto:
> Going back to the original discussion of this policy. `Chat 5.1` is currently contradictory to our current stance with `Common Controls 11.1`. If [Common Controls 11.1](https://github.com/cisagov/ScubaGoggles/blob/main/baselines/Common%20Controls%20Minimum%20Viable%20Secure%20Configuration%20Baseline%20v0.2.md#gwscommoncontrols111v02) allowlisting is not enforced, then turning this setting on allows any user to install any chat app. Which is insecure.
> 
> It seems that we should either delete this baseline and enforce Common Controls as a SHALL but leave the decision to enable this setting up to organizations/agencies
> 
> OR
> 
> We should rewrite this policy to make it organizational unit specific. example:
> 
> ```
> The ability to install Chat apps SHALL be disabled for the top-level organizational unit
> 
> _Note:_ The ability to install Chat apps MAY be enabled on a per OU basis.
> ```
> 
> This is to prevent the default of allowing anyone to install Chat apps if App allowlisting is not enforce but allow flexibility for specific organizational units to install chat applications.

I agree with David on this one. I'll also add that per the research you documented [here](https://github.com/cisagov/ScubaGoggles/issues/222#issuecomment-2037786328):
- "If the new feature is turned on, but the app IS NOT allowlisted, then only admins are allowed to install the app for the org." **Acceptable from a security standpoint.**
- "If the feature is turned on and the app IS allowlisted, then both admins and users can install." **Acceptable from a security standpoint.**
- "If the new feature is off, then the app cannot be installed." **Also acceptable from a security standpoint.**

Therefore, my vote would be that we delete CHAT.5.1 entirely.

jkaufman-mitre:
@adhilto @buidav I can tell neither of you have actually research this setting at all. Turning this setting on does not override Common Controls 11.1. In the contrary, turning this off does not allow users to install any chat app. Turning this feature one allows users to install allowlisted chat apps.

In the future, please make sure to understand the setting before commenting.

jkaufman-mitre:
My vote would be to make this policy a SHOULD because if there is no need for users to have chat apps, then it should be disabled to protect the org.

jkaufman-mitre:
@adhilto, Also I realized that the PR changes did not reflect what my recommendation was, so went in ad commited the updated policy statement.

adhilto:
> @adhilto, Also I realized that the PR changes did not reflect what my recommendation was, so went in ad commited the updated policy statement.

Thanks for pointing that out, that completely changes the discussion.

> @adhilto @buidav Turning this setting on does not override Common Controls 11.1. 

David did not claim that it does. His point was that if Common Controls 11.1 was _not_ implemented and this setting were on, users could install any Chat app. I haven't tested this myself but it certainly seems like that would be the case. Regardless, since you updated it from "SHOULD be enabled" to "SHOULD be disabled," this point is no longer important.

> Not allowing any user to install any app could be a hinderance to the user.

Hindrance, yes. Security risk? No. Though it is a reason for why we wouldn't want to require this setting be disabled, which is now the current recommendation.

Weighing both sides of the issue, the reason for including 5.1 "SHOULD be disabled" is as follows:
- Defense in depth. Yes, COMMON.CONTROLS.11.1 would prevent non-approved apps from being installed (per your research [here ](https://github.com/cisagov/ScubaGoggles/issues/222#issuecomment-2037786328)), but to David's point, what if COMMON.CONTROLS.11.1 is _not_ implemented? Disabling the setting in question for CHAT.5.1 would ensure that even in that case, unapproved Chat apps could not be installed.

Reason for enabling "Allow users to install Chat apps":
- The usability cost in setting that to OFF is quite high, because as you indicated [here](https://github.com/cisagov/ScubaGoggles/issues/222#issuecomment-2037786328), this would completely prevent all Chat apps from being installed, even if the app is allow listed.

Weighing both those two points, I'm inclined to think that this is a setting that we need don't need to take a stance on. Since the security benefit is so minimal and the usability cost so high, each agency should be permitted to weigh the tradeoffs themselves and make their own decision. In other words, I'm still in favor of cutting this control.

That said, as the interactions between these two settings isn't immediately obvious, it could be useful to document and explain this somewhere, so that agencies are able to make an informed decision. We should not mandate that this setting be ON or OFF though.

jkaufman-mitre:
We will go ahead an remove CHAT.5.1v0.2 and make COMMONCONTROLS.11.1v0.2 a SHALL.

jkaufman-mitre:
Policy Group 5 has been removed. @adhilto Ready for review again.

adhilto:
> Policy Group 5 has been removed. @adhilto Ready for review again.

Just need to update the table of contents. Other than that looks good.

adhilto:
@buidav I just implemented the Rego changes, ready for your review.

jkaufman-mitre:
@adhilto @buidav I have updated the Table of Content.

adhilto (approved):

mdueltgen (approved):

mdueltgen:
@adhilto @buidav Any finals thoughts on this or are we good to merge? I think all changes requested were handled.

PR #499, opened 2024-11-14
Policy API Integration

### LINKED ISSUE #83: Find the minimum privileges that need to be assigned to a custom admin role to run the tool
The permissions needed to access the API with the scopes that we need are a bit vague. 
[The reports API Google Documentation guide](https://developers.google.com/admin-sdk/reports/v1/guides/prerequisites) says that a `super admin` or a `custom admin` is needed to access the API. 

Lessons learned from M365, members of the public aren't comfortable with running some random tool off the internet as the highest privileged role in their Cloud environment. For GWS, this is the `super admin` role.

There is no specific Google Documentation for assigning the custom admin the minimum permissions we need to access the reports and directory apis:

This issue is to find out and document the minimum privileges that need to be assigned to a `custom admin` to run this tool.
Then test if there are any issues running the tool as an account assigned just that custom admin role.
[How to create a custom admin role](https://apps.google.com/supportwidget/articlehome?article_url=https%3A%2F%2Fsupport.google.com%2Fa%2Fanswer%2F2406043%3Ftopic%3D2785005&assistant_event=welcome&assistant_id=usermasterbot&product_context=2406043&product_name=UnuFlow&trigger_context=a).

See the README for the OAuth scopes we're currently using for Goggles

### LINKED ISSUE #305: Remove Version Numbers from Control IDs in the Rego
# 💡 Summary #
Remove the version number from the control IDs in the Rego for maintainability.

## Motivation and context ##
The Rego currently references the baseline controls by their full ID, including the version number. For example:
image.

However, the inclusion of the version number makes it challenging to maintain consistency, as the version number is repeated again and again throughout the files. For example, consider https://github.com/cisagov/ScubaGoggles/pull/303, with 101 files changed, just to bump the version number.

Furthermore, moving forward, eventually changing the version numbers will not be a matter of a simple find and replace, as after the 1.0 release, the version numbers for the controls will be independent of each other.

Finally, having the version number listed in the Rego code doesn't really add anything.

## Implementation notes ##

Please provide details for implementation, such as:

- an example for how this would be used
- what this would look like
- how this would act
- any related work, including links to related issues

## Acceptance criteria ##
As a proof of concept, I removed the version numbers from the Sites Rego, like so:
image

Then modified md_parser.py like so:
image

And finally modified reporter.py like so:
image

Worked like a charm:
image

How do we know when this work is done?
- [ ] The version numbers are removed from the Rego
- [ ] The HTML report still looks as expected
- [ ] The json version of the report still looks as expected
- [ ] All unit tests still pass

### LINKED ISSUE #360: Move the baselines and Rego code to within the scubagoggles folder
# 💡 Summary #

This work is to move the Rego code, OPA download script, and baseline folders to within the scubagoggles folder.

## Motivation and context ##

This is necessary if we wish to publish ScubaGoggles on [PyPI](https://pypi.org/)

## Implementation notes ##

Move and refactor all code, links, and references to those files to within the scubagoggles folder.

## Acceptance criteria ##

ScubaGoggles is functional with those folders and files moved within the main ScubaGoggles package directory.

### LINKED ISSUE #412: Add check to ensure OPA executable is available
# 💡 Summary #
Before attempting to execute OPA, first check to ensure the executable is available and at least the minimum version, and provide a helpful error message if not.

## Motivation and context ##
If you try running ScubaGoggles without the executable, the error message does not provide the user with actionable information.

## Implementation notes ##
See https://github.com/cisagov/ScubaGoggles/issues/410.

## Acceptance criteria ##
- [ ] Check to see if executable is present is implemented
- [ ] Check to see if acceptable version is implemented
- [ ] Error message contains actionable information

### LINKED ISSUE #430: Correct pylint Errors
# 💡 Summary #
Correct the following lint errors:
image

## Motivation and context ##
pylint version 3.3.0, which was just released (Sep 20, 2024), apparently added checks that our code does not currently pass. As a temporary measure I've pinned pylint in the workflow to version 3.2.7. We should update the workflow to the latest version and either refactor our code to pass the checks or configure more exceptions in the workflow.

## Implementation notes ##
pylint version pinned here: https://github.com/cisagov/ScubaGoggles/blob/c8d0c9357792f569fb18bcae4cb56f99843af321/.github/workflows/pylint.yml#L34

## Acceptance criteria ##
- [ ] Workflow uses the latest version of pylint
- [ ] Workflow passes

### LINKED ISSUE #449: Invert the `--omitsudo` flag
# 💡 Summary #
Either:
- Completely remove the built-in functionality to run the OPA executable with sudo
- Change the `--omitsudo` flag into a `--runwithsudo` flag so that it defaults to running it without sudo

## Motivation and context ##
Currently, ScubaGoggles by default runs the OPA executable as sudo on Linux and Mac. Previous tests indicated that running it as sudo was necessary, but latest tests indicate that it can be run without sudo. Running it with sudo by default introduces security risks that we should avoid if possible.

## Implementation notes ##
If we do keep the built-in functionality to run OPA with sudo, we might consider adding a last-second hash check to minimize the security risk.

## Acceptance criteria ##
- [ ] OPA is no longer run with sudo by default

### LINKED ISSUE #465: GWS.COMMONCONTROLS.11.2v0.3 does not detect setting inheritance
# 🐛 Summary #
GWS.COMMONCONTROLS.11.2v0.3 does not detect setting inheritance (for both OUs and groups)

## To reproduce ##
1. Follow https://github.com/cisagov/ScubaGoggles/blob/main/baselines/commoncontrols.md#gwscommoncontrols112v03-instructions. Ensure there is an event for the top-level OU
2. Make a child OU or a group non-compliant.
3. Click "Inherit" (or "Unset" for a group).
4. Run ScubaGoggles. The child OU or group will still show as non-compliant.

## Expected behavior ##
The child OU / group to show as compliant if it inherits from its parent.

## Any helpful log output or screenshots ##
N/A

### LINKED ISSUE #498: ScubaGoggles with Policy API integration
This is the complete integration of the ScubaGoggles version containing the Google Policy API implementation.

rlxdev:
# ScubaGoggles with Google's Policy API Implementation

This is the complete integration of the version of ScubaGoggles containing the
implementation of Googles's Policy API.

- All Rego Tests Pass, including **185 new tests** covering the baselines
  implemented by the Policy API.
- Regal runs on all Rego code with no issues.
- PYLint runs on all Python code with no issues.
- Mitchel's "smoke" tests pass (locally, after modifying for credentials &
  scubagws tenant).
- This version has been tested on Windows 10, with varying Python versions,
  and RedHat RHEL 8 linux.

### Google Policy API

The Google Policy API is used for all baselines that can be assessed by the data
returned by the API.  Google returns all settings available for the top-level
orgunit, and then only the settings that differ from the top-level orgunit for
sub-orgunits and groups.

The Policy API data is available to the rego code in the input data in the
"policies" section.  The settings are grouped by orgunit or orgunit/group name.
There is **no** distinction between orgunits and groups.

The Policy API is queried using Google's `AuthorizedSession` class because
access to the API is not yet available using the Google API client, which is
used for the other APIs queried by ScubaGoggles (reports, directory, groups).
There have been **no** issues using this method to access the Policy API.

Because of the way Rego handles undefined variables, basically returning
`undefined` when a variable is not defined, the Python code fetching the Policy
API data **must** verify that all expected settings are present in the top-level
orgunit's data.  It's expected that missing settings will be rare, but they
can't be ignored because missing settings will lead to incorrect results.
ScubaGoggles doesn't have to abort on any missing settings, so currently a
warning is issued for any missing settings.

As with the log events, the Rego code does the work to determine whether the
requirement has been met for each baseline covered by the Policy API.  In
implementing the Policy API in Rego, the log event code has been kept along
with the new Policy API code for debugging/testing during the development of
the Policy API implementation.  This code will be removed in a subsequent
release.

### ScubaGoggles Package Distribution & Installation

This version of ScubaGoggles is able to be packaged using the standard Python
packaging software, using the `setuptools` back-end.  Being able to install
ScubaGoggles and its dependencies as a package using `pip` makes the
installation much easier for users and provides a separation between the
distributed code and the user's data (credentials and reports).

Unfortunately, the current configuration in the repository does not conform to a
structure that can be packaged.  This has led to subtle issues, such as circular
import errors when running ScubaGoggles, and the incorrect placement of
ScubaGoggles-related directories in the top-level `site-packages` directory in
the Python environment.

For this version, directories that are included in the package have been
moved below the `scubagoggles` directory:

- `baselines`
- `docs`
- `rego`
- `sample-config-files`
- `sample-report`
- `Testing`
- `utils`

The current use of `setup.py`, with package configuration included in the file,
is deprecated by `setuptools`.  This version now uses `pyproject.toml` for
package configuration.

The package is built using the `scubagoggles/utils/build.sh` script.  It
produces a "wheel" file (`.whl`) that is installable using `pip`.

In addition to installing ScubaGoggles, users have to specify:

- location where the reports will be created.
- Google OAuth credentials file (`credentials.json`).
- location of the OPA executable.

For this version, users will run `scubagoggles setup` to provide those
locations.  Unless the user explicitly changes a location, the setup is a
one-time step, even after software updates.

What are the advantages of making ScubaGoggles an installable package:

- Package installation is easier for users.  They will not have to install
  dependent packages separately.
- User's will not have to change their working directory to run ScubaGoggles.
- The `scubagoggles` command will always work and there won't be a need to have
  them try `Python scuba.py`.
- Upgrading ScubaGoggles will be easy and can be done in an existing (virtual)
  environment.  User's will not have to relocate files, such as the
  `credentials.json` or OPA executable.

### Versioning

There are two forms of the version number used for ScubaGoggles: the software
version number (e.g., "0.3.0"), and what I call the "version suffix" that
proliferates throughout the files referencing baseline identifiers.  There are
multiple places in the code that define the version number.  Whenever there is a
version upgrade, the change touches nearly every file in the repository.

For this version, the "official" version number is defined in only **one** place
(`scubagoggles/__init__.py`). All other references to the version number are
derived from this value.  The version number is managed using the `scubagoggles
version` command.

The Python code **must** use the `Version` class when referencing the version
number.  For the Rego code, the version suffix used in the baseline identifiers
is accessible in the input written by the Python code.

Baseline identifiers in the Rego code are defined once in a variable for each
baseline, and all references to the identifier are to use this variable
(including the test Rego code).  For example:

```
CommonControlsId3_2 := utils.PolicyIdWithSuffix("GWS.COMMONCONTROLS.3.2")
```

The `utils.PolicyIdWithSuffix()` function handles the addition of the version
suffix at the end of the identifier.  There **must** be no reference to the
version suffix in any comments in the Rego code (in fact, there should be
**no** occurrences of the hard-coded version suffix in any Rego file).  It
is also not necessary to reference a specific version suffix in the user
documentation.

The "drift rules" and baseline Markdown files will continue to have the
version suffix included in the baseline version.  The version suffixes in
these files will be modified as part of the version upgrade process.

When a version upgrade occurs, these are the only files that are modified
in the process:

- `README.md`
- `drift-rules/GWS Drift Monitoring Rules - Calendar.csv`
- `drift-rules/GWS Drift Monitoring Rules - Chat.csv`
- `drift-rules/GWS Drift Monitoring Rules - Classroom.csv`
- `drift-rules/GWS Drift Monitoring Rules - Common Controls as of 11-14-23.csv`
- `drift-rules/GWS Drift Monitoring Rules - Drive and Docs.csv`
- `drift-rules/GWS Drift Monitoring Rules - Gmail.csv`
- `drift-rules/GWS Drift Monitoring Rules - Groups.csv`
- `drift-rules/GWS Drift Monitoring Rules - Meet.csv`
- `drift-rules/GWS Drift Monitoring Rules - Sites.csv`
- `scubagoggles/__init__.py`
- `scubagoggles/baselines/calendar.md`
- `scubagoggles/baselines/chat.md`
- `scubagoggles/baselines/classroom.md`
- `scubagoggles/baselines/commoncontrols.md`
- `scubagoggles/baselines/drive.md`
- `scubagoggles/baselines/gmail.md`
- `scubagoggles/baselines/groups.md`
- `scubagoggles/baselines/meet.md`
- `scubagoggles/baselines/sites.md`

### Pandas Dependency Dropped

The pandas module was included as a requirement because it was used to create
an HTML table for the reports.  Pandas is a "beast" of a module, and also
requires another "beast" in numpy.  The table creation is easy in this case,
and it allows us to drop a module that really isn't being used for its
intended purpose (data analysis).  Having pandas dropped as a requirement also
lessens the potential issues users will have when installing ScubaGoggles.

### No "sudo" for Running OPA

There was some issue with OPA in the past that apparently required the use of
"sudo" (privilege escalation) to run it on macOS and perhaps linux.  This
issue seems to be fixed, so the ability to use "sudo" for running OPA has
been removed.  OPA is basically a language interpreter like Python - they are
user mode programs that should never require privilege escalation.  If this
happens again, the onus is on OPA to fix the issue, rather than trying to
work-around it.

### Python Logging Facility Enabled

The Python logging facility has been enabled, so logging may be inserted into
the code where appropriate using `log.{debug, info, warning, error, critical}()`
function calls.  The current default level that is displayed to the user is
`warning` (meaning that `warning` level messages and above will be displayed and
not lower level (`debug, info`)), but this can be altered on the command line
using the `--log` option.

### Downloading OPA

With users installing ScubaGoggles as a Python package, they can't easily
run the `download_opa.py` script.  To continue to provide users the ability
to download OPA, the `scubagoggles getopa` command performs the same task, but
also ensures the executable has the correct permissions to run on macOS and
linux.

### Ability to Purge Output Directories

This might be more useful for developers running ScubaGoggles often for testing,
but there's now the ability via `scubagoggles purge` to delete older versions
of the reports output directories.

Closes #498
Closes #449
Closes #360
Closes #430
Closes #412
Closes #305
Closes #465
Closes #83

### Follow-up issues
- #502
- #503
- #504 
- #505 
- #506 
- #507 
- #508 
- #509 
- #510 
- #511 
- #512
- #513

## 🧪 Testing

These changes have been continuously tested during the implementation of the Policy API settings released by Google.
As mentioned in the description, all automated tests (PYLint, regal, Rego tests) pass.

## ✅ Pre-merge Checklist

- [x] Squash all commits into one PR level commit using the `Squash and merge` button.

## ✅ Post-merge Checklist


adhilto:
Tagging @mdueltgen @jkaufman-mitre for awareness, this PR contains baseline changes. No substantive changes to the policies, just changes along the line of changing the policies section headers to "Policies" for consistency.

baselines/README.md
adhilto, line 1: Create a placeholder "baselines" folder at the top-level of the repo, with a README linking to the actual baselines
folder.
- See https://github.com/cisagov/ScubaGear/tree/main/baselines for reference.
- CISA's website refers users to GitHub to view the baslines, we want it to be easy for users to find the baselines
- Some third-party websites already link to that folder. If we put a README there those links won't break. For example: https://www.googblogs.com/page/12/

docs/authentication/OAuth.md
adhilto, line 1: Recommend moving the docs folder back.
Move the docs folder back
- Some existing CISA documentation assumes the docs folder is at the top-level of the repo
- Consistency with ScubaGear
- Would break GitHub pages integration should we want to implement that (ScubaGear has: https://cisagov.github.io/ScubaGear/)

scubagoggles/docs/authentication/OAuth.md
adhilto, line 1: Recommend removing screenshots from this file.
- An early decision made on SCuBA was to not include screenshoots of the admin centers, due to the highly fluid nature
of SaaS products.
- GitHub does not render the screenshots well, the formatting is very skewed.

scubagoggles/user_setup.py
adhilto, line 1: This needs to be reworked. With this update, ScubaGoggles now takes two conflicting config files, each playing different roles.
The ".scubagoggles" config created by the setup utility and the config file users can create and indicate using the
`-config` parameter. The .scubagoggles config only takes `credentials`, `opa_dir`, and `output_dir`. The config file
pointed to by the `-config` parameter can specify all CLI args (including those three locations, but using the names
spelled out by the cli: `credentials`, `opapath `, and `outputpath`).
- First, the `.scubagoggles` config parameters should match the parameters used for the CLI.
- Second, why do we need the `.scubaconfig` config file at all? All the features it covers are already provided by the
config file option already built into ScubaGoggles.
I'd recommend removing this feature altogether as the funtionality of `.scubagoggles` is already built in to ScubaGoggles.

I understand the desire to improve the user experience, but this misses the mark.

My recommendations:
- Short term: cut this feature from ScubaGoggles
- Long term: After this PR is merged in, we discuss as a team what approach we want to take to improve the setup experience, then implement whatever approach we decide on as a team.

scubagoggles/scuba_argument_parser.py
adhilto, line 1: Fix specifying "credentials" via config file. Now, if you specify `credentials` in a config file you get a error like this:
```
args.credentials = args.credentials.resolve()
AttributeError: 'str' object has no attribute 'resolve'
```

scubagoggles/auth.py
```diff
@@ -4,54 +4,167 @@
...
+
+    _scopes = (f'{_base_auth_url}/admin.reports.audit.readonly',
+               f'{_base_auth_url}/admin.directory.domain.readonly',
+               f'{_base_auth_url}/admin.directory.orgunit.readonly',
+               f'{_base_auth_url}/admin.directory.user.readonly',
+               f'{_base_auth_url}/admin.directory.group.readonly',
+               f'{_base_auth_url}/apps.groups.settings',
+               f'{_base_auth_url}/cloud-identity.policies.readonly')
+
+    def __init__(self, credentials_path: Path, svc_account_email: str = None):
```
adhilto, line 37: Fix service account authentication:
- The `GwsAuth` determines if the creds file is for a service account by checking if the `svc_account_email` parameter is not `None`
- The `GwsAuth` has that parameter default to `None`
- The provider instantiates the auth class like so: `GwsAuth(credentials_file)`, meaning `svc_account_email` will always be `None`, even if the user provides one.
- This causes the auth class to improperly handle the service account credentials, resulting in `ValueError: Client secrets must be for a web or installed app.`

scubagoggles/docs/prerequisites/Prerequisites.md
```diff
@@ -1,25 +1,25 @@
...
+
+The tool uses the following OAUTH API scopes:
+
+```
+https://www.googleapis.com/auth/admin.reports.audit.readonly,
+https://www.googleapis.com/auth/admin.directory.domain.readonly,
+https://www.googleapis.com/auth/admin.directory.group.readonly,
+https://www.googleapis.com/auth/admin.directory.orgunit.readonly,
+https://www.googleapis.com/auth/admin.directory.user.readonly,
+https://www.googleapis.com/auth/apps.groups.settings
```
adhilto, line 13: Add missing scope. This is particularly important for service account users as we instruct the users to copy and paste that list verbatim when authorizing the service account.

```suggestion
https://www.googleapis.com/auth/apps.groups.settings,
https://www.googleapis.com/auth/cloud-identity.policies.readonly
```

scubagoggles/main.py
```diff
@@ -6,119 +6,399 @@
...
+
+    parser.set_defaults(dispatch = getopa)
+
+    parser.set_defaults(user_config = user_config)
+
+    parser.add_argument('--check',
+                        '-c',
+                        default = False,
+                        action = 'store_true',
+                        help = 'Check hash code after download')
```
adhilto, line 224: Change this to something like "skipcheck", defaulting to False. Following secure by design principles, the secure option needs to be the default.

adhilto (changes requested):
I've barely scratched the surface of the testing that needs to be done, but here's a first round of feedback.

Two types of feedback:
- Change now
- Change soon but after this PR.

I've left individual comments on changes that should be made before this PR is merged. 
I'm listing the changes that can wait here in this comment. In the spirit of keeping the scope of this PR from expanding any further, these changes can wait but should be tracked as issues before this PR is merged.

**Simplify the rego code**
Controls that have policy API support should exclusively use the policy API. No need to maintain the old reports code. Would greatly simplify the code. (I know you note this in the description but we still need to track it as an issue).

**Rework the baseline version logic.**
This code assumes that all controls will have the same version number. This is not a safe assumption. Pre-version 1.0, this will always be the case. But after the 1.0 release, if a control needs to be updated, only the version number of that control will be updated. For reference, see the [Exchange Online baseline](https://github.com/cisagov/ScubaGear/blob/main/PowerShell/ScubaGear/baselines/exo.md). Note that not all controls have the same version number (e.g., MS.EXO.1.1v1 and MS.EXO.2.2v2).

**Update the "prerequisites" field of the rego tests**
As noted in a comment in the reporter, "If Prerequisites is not defined, assume the test just depends on the reports API." This is no longer a safe assumption.

**Update the report details message for controls relying on log events**
Clearly indicate the controls that can only be checked via the log events. These checks give imperfect visibility which should be highlighted to the users. For example, if the log events for the top level OU is show compliance, ScubaGoggles will currently report "Requirement met for all OUs and groups," but in reality we cannot know that with full confidence.

scubagoggles/user_setup.py
rlxdev: I don't see this as conflicting.  The one I've implemented is basically a one-time configuration for the user to provide the 3 things we need from them in order to run: the output directory location, the OPA executable directory, and the location of the credentials file.  The other configuration is basically a run-time configuration that alters how ScubaGoggles runs, and I can see that the user may have more than one of those configuration files in use for different runs.

The user does not even need to really know about the "tool" configuration file and the intent was for it to be modified by the `setup` command.  These settings apply across runs of ScubaGoggles.  It's possible that the user might want to change the output directory location, but in practice I think it's not that likely but it can be done on the command line.

This does simplify the user's initial setup and I don't see a conflict with the baseline configuration, so I don't see the need to change this.

scubagoggles/docs/authentication/OAuth.md
rlxdev: That's fine, but I'm seeing from observing user sessions that this is a confusing part of the setup.  Obviously, a picture is worth 1000 words, but if we are taking out the screenshots we need to really make this as easy as possible for the user to understand.

docs/authentication/OAuth.md
rlxdev: That's OK.  I did it originally thinking I'd get push back on not having it included with the distributed package.  If that's OK, we can put it back.

rlxdev:
> **Simplify the rego code** Controls that have policy API support should exclusively use the policy API. No need to maintain the old reports code. Would greatly simplify the code. (I know you note this in the description but we still need to track it as an issue).

This has been the plan all along.  I approached this conservatively and preserved the log event code to keep the testing working.
My thought is to remove this in the 1.1 version, and it can be done in a methodical way throughout.

> **Rework the baseline version logic.** This code assumes that all controls will have the same version number. This is not a safe assumption. Pre-version 1.0, this will always be the case. But after the 1.0 release, if a control needs to be updated, only the version number of that control will be updated. For reference, see the [Exchange Online baseline](https://github.com/cisagov/ScubaGear/blob/main/PowerShell/ScubaGear/baselines/exo.md). Note that not all controls have the same version number (e.g., MS.EXO.1.1v1 and MS.EXO.2.2v2).

The way I've implemented the "version suffix" changes doesn't preclude the possibility of having different suffixes for certain baseline policies.  When we do end up having different suffixes on the baselines, it will not require much additional coding.  It must be implemented such that there is only **one** source of the baseline version, most likely the markdown file where the baseline is defined.  The correct suffix for the policies can be obtained from the source and driven across the system through the code instead of hard-coding the suffix in multiple places.

scubagoggles/docs/prerequisites/Prerequisites.md
```diff
@@ -1,25 +1,25 @@
...
+```
+
+When running ScubaGoggles for the first time you will be prompted to consent to these API scopes. Users with the Super Admin role automatically have the privilege to consent to these scopes. A custom admin role can also be made with the minimum permissions to consent to these scopes. See this [Google Admin SDK Prerequisites guide](https://developers.google.com/admin-sdk/reports/v1/guides/prerequisites) for more information.
```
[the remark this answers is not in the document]
rlxdev: done

scubagoggles/docs/usage/Config.md
```diff
@@ -1,39 +1,39 @@
...
-- Return to [Documentation Home](/README.md)
+
+# Usage: Config File
+All ScubaGoggles [parameters](/docs/usage/Parameters.md) can be placed into a configuration file in order to made execution easier. The path of the file is specified by the `--config` parameter, and its contents are expected as YAML.
+
+> [!NOTE]
+> If a parameter is specified both on the command-line and in a configuration file, the command-line parameter has precedence over the config file.
+
+## Sample Configuration Files
+[Sample config files](/sample-config-files) are available in the repo and are discussed below. When executing ScubaGoggles, only a single config file can be read in; we recommend looking through the following examples and constructing a config file that best suits your use case.
```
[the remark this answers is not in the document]
rlxdev: done

scubagoggles/docs/usage/Config.md
```diff
@@ -1,39 +1,39 @@
...
+scubagoggles gws --config basic_config.yaml -b gmail chat
+```
+
+### Omit Policies
+
+In some cases, it may be appropriate to omit specific policies from ScubaGoggles evaluation. For example:
+- When a policy is implemented by a third-party service that ScubaGoggles does not audit.
+- When a policy is not applicable to your organization (e.g., policy GWS.GMAIL.4.3, which is only applicable to federal, executive branch, departments and agencies).
+
+The `omitpolicy` top-level key, shown in this [example ScubaGoggles configuration file](/sample-config-files/omit_policies.yaml), allows the user to specify the policies that should be omitted from the ScubaGoggles report. Omitted policies will show up as "Omitted" in the HTML report and will be colored gray. Omitting policies must only be done if the omissions are approved within an organization's security risk management process. **Exercise care when omitting policies because this can inadvertently introduce blind spots when assessing your system.**
```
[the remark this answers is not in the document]
rlxdev: done

scubagoggles/scuba_argument_parser.py
rlxdev: done

scubagoggles/auth.py
rlxdev: done - this should work - I don't have access to a service account to test.

scubagoggles/docs/prerequisites/Prerequisites.md
rlxdev: done

scubagoggles/main.py
rlxdev: done

scubagoggles/docs/usage/Examples.md
```diff
@@ -1,49 +1,49 @@
...
-> [!NOTE]
-> In all the above examples, the html report should open automatically. If not, navigate to the output folder and open the `*.html` file using a browser of your choice. The json output will also be located in this folder.
-
-## Navigation
-- Continue to [Reviewing Output](/docs/usage/ReviewOutput.md)
-- Return to [Documentation Home](/README.md)
+
+# Usage: Examples
+> [!Note] (This is intended for developers ONLY)
+> If you chose not install the `scubagoggles` package in a venv but do have the dependencies installed from `requirements.txt`, you may execute the tool using the `scuba.py` script located in the root directory of this repository. Replace any `scubagoggles` directions with `python scuba.py`
```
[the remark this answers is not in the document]
adhilto: We could probably just cut this note entirely, especially if we don't want users to invoke ScubaGoggles this way.

docs/authentication/OAuth.md
adhilto: That's ok. ScubaGear takes the same approach.

scubagoggles/auth.py
adhilto: I can confirm, service account authentication is working now, thanks for the updates.

scubagoggles/scuba_argument_parser.py
```diff
@@ -1,137 +1,170 @@
...
+                      + Version.suffix)
+            policy_id_map[policy_id] = new_id
+
+        args.omitpolicy = {new_id: omit_policies[old_id]
+                           for old_id, new_id in policy_id_map.items()}
+
+        return
+
+    @staticmethod
+    def validate_config(args : argparse.Namespace) -> None:
```
adhilto, line 132: Specifying `opapath` in a config file now gives `AttributeError: 'str' object has no attribute 'resolve'`

adhilto (changes requested):

scubagoggles/user_setup.py
adhilto: The approach I would recommend we take:

The setup utility creates a .scubagoggles folder in the user's home directory
The setup utility automatically downloads the OPA executable and stores it in the .scubagoggles folder. `
~/.scubagoggles is made the default OPA location.
Further notes:

When I say the setup utility should automatically download the OPA executable, I'm not saying we should remove the separate scubagoggles getopa function, users should still be able to call that function directly if desired. But if the setup utility automatically invokes the getopa function, that's one less thing for the user to worry about.
We could make the setup utility also create or ask the user to specify a default output folder, but I don't see that as essential. If we did that then yes, we still would need a .scubagoggles file, but within the .scubagoggles folder.
We could also make the setup utility ask for the location of the user's credentials file, but I think that should be optional, as some users might have multiple organizations they manage and would prefer to specify which credentials to use each time. Either way, if the setup utility is going to ask the user for the location of the credentials file, it should do so after the user has created it (currently the README tells them to run the setup utility first, which I think will confuse users, since it asks them for a file they don't have yet).

README.md
```diff
@@ -29,42 +33,47 @@ We use a three-step process:
 3. **Report**. Package the results as HTML and JSON.
 
 ## Limitations of the tool
-The majority of the conformance checks done by ScubaGoggles rely on [GWS Admin log events](https://support.google.com/a/answer/4579579?hl=en). If there is no log event corresponding to a SCuBA baseline policy, ScubaGoggles will indicate that the setting currently can not be checked on its HTML report output. In this situation, we recommend you manually review your GWS security configurations with the SCuBA security baselines. See [Limitations](/docs/usage/Limitations.md) for more details.
+The majority of the conformance checks done by ScubaGoggles rely on
```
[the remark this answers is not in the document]
rlxdev: done

docs/prerequisites/Prerequisites.md
```diff
@@ -1,25 +1,30 @@
...
+https://www.googleapis.com/auth/admin.directory.user.readonly,
+https://www.googleapis.com/auth/apps.groups.settings,
+https://www.googleapis.com/auth/cloud-identity.policies.readonly
+```
+
+When running ScubaGoggles for the first time you will be prompted to consent to
+these API scopes. Users with the Super Admin role automatically have the
+privilege to consent to these scopes. When running ScubaGoggles for the first
+time you will be prompted to consent to these API scopes. Users with the Super
+Admin role automatically have the privilege to consent to these scopes.
```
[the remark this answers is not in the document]
rlxdev: done

scubagoggles/scuba_argument_parser.py
rlxdev: done

snarve (commented):
In the Testing section, it would help to have instructions on how to update any dependencies (automatically by running specific scripts/commands) or download any required packages.
Since this PR covers major updates, assuming running the tool now requires a different command (or atleast different flags), listing the instructions out in the Testing section would help.

image

scubagoggles/Testing/Functional/SmokeTests/smoke_test.py
```diff
@@ -0,0 +1,82 @@
...
+
+        Args:
+            subjectemail: (optional) The email address of a user for the
+            service account.  If not specified, the user's OAuth credentials
+            are used instead.
+        """
+
+        svc_account_option = (f' --subjectemail {subjectemail}' if subjectemail
+                              else '')
+        command = f'scubagoggles gws{svc_account_option} --quiet'
```
adhilto, line 46: The subjectemail is not optional for the smoke test, and for some reason including the `= None` made it so that subjectemail would always be None, even if you specified one to pytest. I tested locally, the following change fixed the smoke test for me (tested on Windows).

```suggestion
    def test_scubagoggles_output(subjectemail: str):
        """
        Test if the `scubagoggles gws` command generates correct output for
        all baselines.

        Args:
            subjectemail: The email address of a user for the
            service account.
        """

        command = f'scubagoggles gws --subjectemail {subjectemail} --quiet'
```

scubagoggles/Testing/Functional/SmokeTests/README.md
adhilto, line 1: The paths in this README have not been updated to reflect the new testing folder location.

adhilto (changes requested):

scubagoggles/Testing/Functional/SmokeTests/smoke_test.py
rlxdev: Because we don’t all have access to a service account, this smoke test must be able to be run with either the service account or the OAuth credentials.  I thought the default value would work correctly (but I couldn’t test it because I don’t have access to a service account).  I’m going to try a different tack with this change and hopefully it’ll work with the service account.

scubagoggles/rego/Drive.rego
```diff
@@ -0,0 +1,1588 @@
...
+if {
+    Check1_2_OK
+    Conditions := {count(NonCompliantOUs1_2) == 0, count(NonCompliantGroups1_2) == 0 }
+    Status := (false in Conditions) == false
+}
+#--
+
+#
+# Baseline GWS.DRIVEDOCS.1.3
+#--
```
adhilto, line 222: If you select "ALLOWLISTED DOMAINS" and don't check "Warn when files owned by users or shared drives in...are shared with users in allowlisted domains," 1.3 still passes.

scubagoggles/rego/Drive.rego
```diff
@@ -0,0 +1,1588 @@
...
+    Check1_3_OK
+    Conditions := {count(NonCompliantOUs1_3) == 0, count(NonCompliantGroups1_3) == 0}
+    Status := (false in Conditions) == false
+}
+
+#--
+
+#
+# Baseline GWS.DRIVEDOCS.1.4
+#--
```
adhilto, line 321: If you select "ALLOWLISTED DOMAINS" and check "Allow users or shared drives in... to share items with non-Google users in trusted domains using visitor sharing" 1.4 still passes.

If you select "ON" and check "Allow users or shared drives in...to share items with people outside...who aren't using a Google account," 1.4 fails, as it should, but the report details string is incorrect: "External Sharing is enabled but sharing items to non-google accounts is disabled" ("disabled" instead of "enabled").

scubagoggles/rego/Drive.rego
```diff
@@ -0,0 +1,1588 @@
...
+    Check1_4_OK
+    Conditions := {count(NonCompliantOUs1_4) == 0, count(NonCompliantGroups1_4) == 0}
+    Status := (false in Conditions) == false
+}
+
+#--
+
+#
+# Baseline GWS.DRIVEDOCS.1.5
+#--
```
adhilto, line 446: If you have "When sharing outside...is allowed, users...can make files and published web content visible to anyone with the link" selected, then set outside sharing to OFF, 1.5 still fails.

scubagoggles/rego/Drive.rego
```diff
@@ -0,0 +1,1588 @@
...
+if {
+    Check1_6_OK
+    Conditions := {count(NonCompliantOUs1_6) == 0, count(NonCompliantGroups1_6) == 0}
+    Status := (false in Conditions) == false
+}
+#--
+
+#
+# Baseline GWS.DRIVEDOCS.1.7
+#--
```
adhilto, line 626: If you have sharing set to OFF but had "Distributing content outside of..." enabled first, 1.7 still fails.

scubagoggles/rego/Drive.rego
```diff
@@ -0,0 +1,1588 @@
...
+}
+#--
+
+###################
+# GWS.DRIVEDOCS.2 #
+###################
+
+#
+# Baseline GWS.DRIVEDOCS.2.1
+#--
```
adhilto, line 820: The report fails you no matter what you select for "Allow members with manager access to override the settings below"

scubagoggles/rego/Drive.rego
```diff
@@ -0,0 +1,1588 @@
...
+if {
+    Check2_1_OK
+    Conditions := {count(NonCompliantOUs2_1) == 0, count(NonCompliantGroups2_1) == 0 }
+    Status := (false in Conditions) == false
+}
+#--
+
+#
+# Baseline GWS.DRIVEDOCS.2.2
+#--
```
adhilto, line 906: The report fails you no matter what you select for "Allow users outside... to access files in shared drives"

scubagoggles/rego/Drive.rego
```diff
@@ -0,0 +1,1588 @@
...
+if {
+    Check2_2_OK
+    Conditions := {count(NonCompliantOUs2_2) == 0, count(NonCompliantGroups2_2) == 0 }
+    Status := (false in Conditions) == false
+}
+#--
+
+#
+# Baseline GWS.DRIVEDOCS.2.3
+#--
```
adhilto, line 992: The report passes no matter what you select for "Allow people who aren't shared drive members to be added to files"

scubagoggles/rego/Drive.rego
```diff
@@ -0,0 +1,1588 @@
...
+if {
+    Check2_3_OK
+    Conditions := {count(NonCompliantOUs2_3) == 0, count(NonCompliantGroups2_3) == 0 }
+    Status := (false in Conditions) == false
+}
+#--
+
+#
+# Baseline GWS.DRIVEDOCS.2.4
+#--
```
adhilto, line 1078: The report fails you no matter what you select for "Allow viewers and commenters to download, print, and copy files"

scubagoggles/rego/Drive.rego
```diff
@@ -0,0 +1,1588 @@
...
+}
+#--
+
+###################
+# GWS.DRIVEDOCS.3 #
+###################
+
+#
+# Baseline GWS.DRIVEDOCS.3.1
+#--
```
adhilto, line 1169: The report fails you no matter what you select for the security update settings

scubagoggles/rego/Drive.rego
```diff
@@ -0,0 +1,1588 @@
...
+}
+#--
+
+###################
+# GWS.DRIVEDOCS.6 #
+###################
+
+#
+# Baseline GWS.DRIVEDOCS.6.1
+#--
```
adhilto, line 1455: The report fails you no matter what you select for the drive for desktop settings

adhilto (changes requested):

scubagoggles/rego/Drive.rego
```diff
@@ -0,0 +1,1588 @@
...
+
+LogMessage5_1 := "ENABLE_DOCS_ADD_ONS"
+
+Check5_1_OK if {
+    # not PolicyApiInUse
+    events := utils.FilterEventsOU(LogEvents, LogMessage5_1, utils.TopLevelOU)
+    count(events) > 0
+}
+
+Check5_1_OK if {PolicyApiInUse}
```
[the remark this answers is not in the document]
rlxdev: done

scubagoggles/rego/Groups.rego
```diff
@@ -0,0 +1,502 @@
...
+
+Check4_1_OK if {PolicyApiInUse}
+
+GetFriendlyValue4_1(Value) := "Administrators only" if {
+    Value == "ADMIN_ONLY"
+} else := "Users in your domain only" if {
+    Value == "USERS_IN_DOMAIN"
+} else := "Any user" if {
+    Value == "WORLD"
+} else := Value
```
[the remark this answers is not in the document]
rlxdev: done

scubagoggles/rego/Groups.rego
```diff
@@ -0,0 +1,502 @@
...
+    Value == "OWNERS"
+} else := "Managers" if {
+    Value == "MANAGERS"
+} else := "Group members only" if {
+    Value == "GROUP_MEMBERS"
+} else := "Users in your domain only" if {
+    Value == "DOMAIN_USERS"
+} else := "Any user" if {
+    Value == "PUBLIC"
+} else := Value
```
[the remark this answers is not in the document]
rlxdev: done

scubagoggles/rego/Drive.rego
rlxdev: done

scubagoggles/rego/Drive.rego
rlxdev: done

scubagoggles/rego/Drive.rego
rlxdev: done

scubagoggles/rego/Drive.rego
rlxdev: done

scubagoggles/rego/Drive.rego
rlxdev: image

scubagoggles/rego/Drive.rego
rlxdev: image

scubagoggles/rego/Drive.rego
rlxdev: image

scubagoggles/rego/Drive.rego
rlxdev: image

scubagoggles/rego/Drive.rego
rlxdev: This one doesn't have a flag about fixed settings due to the free license, but I suspect this is the case.  The setting for "
Allow users to remove/apply the security update for files they own or manage" is True no matter the state of the checkbox.

image

`'drive_and_docs_file_security_update': {'allowUsersToManageUpdate': True,`

I will check that this might be a fixed setting at the Monday Google meeting.

scubagoggles/rego/Drive.rego
rlxdev: image

adhilto: The org we use for testing uses Google Workspace Enterprise Plus, so that isn't the issue.

scubagoggles/rego/Drive.rego
adhilto: The org we use for testing uses Google Workspace Enterprise Plus, so that isn't the issue.

scubagoggles/rego/Drive.rego
adhilto: The org we use for testing uses Google Workspace Enterprise Plus, so that isn't the issue.

scubagoggles/rego/Commoncontrols.rego
```diff
@@ -504,22 +543,32 @@ NonCompliantOUs3_1 contains {
     LastEvent.NewValue != "DELETE_APPLICATION_SETTING"
 }
 
+NonCompliantOUs3_1 contains {
+    "Name": OU,
+    "Value": NonComplianceMessage3_1
+}
+if {
+    some OU, settings in input.policies
+    enableChallenge := settings.security_login_challenges.enableEmployeeIdChallenge
```
adhilto, line 552: `settings.security_login_challenges.enableEmployeeIdChallenge` isn't actually the setting that corresponds to 3.1. It looks like 3.1 isn't in fact currently supported by the policy API.

scubagoggles/rego/Commoncontrols.rego
```diff
@@ -599,33 +664,58 @@ NonCompliantOUs4_1 contains {
...
 }
 
-tests contains {
-    "PolicyId": "GWS.COMMONCONTROLS.4.1v0.3",
+# The following checks for web session duration less than or equal to 12 hours.
+# Not sure whether this really fits with 4.1 requirement.  The 12 hours is
+# mentioned, but the requirement deals with forced re-authentication after the
+# session expiration - is this automatic with session expiration?  If so, then
+# this check is probably appropriate.  It seems to fit with the event log
+# method above.
```
adhilto, line 672: Recommend trimming this comment - this is the setting relevant to 4.1, no uncertainty about it.
```suggestion
# The following checks for web session duration less than or equal to 12 hours.
```

scubagoggles/rego/Commoncontrols.rego
```diff
@@ -599,33 +664,58 @@ NonCompliantOUs4_1 contains {
...
+NonCompliantOUs4_1 contains {
+    "Name": OU,
+    "Value": sprintf("Web session duration: %s", [duration])
+}
+if {
+    multipliers := {"s": 1, "m": 60, "h": 3600}
+    # This is the requirement limit for session duration:
+    webSessionMax := 12 * multipliers["h"]
+    some OU, settings in input.policies
+    duration := settings.security_session_controls.webSessionDuration
```
adhilto, line 683: This another instance where the policy API returns a fixed value, `1209600s`, or two weeks. But oddly enough _just_ for the primary OU, it seems to correctly handle the child OUs.

adhilto (changes requested):

scubagoggles/rego/Commoncontrols.rego
rlxdev: This did correspond to 3.1:

image

but it was removed a month ago for some reason.  The rego code has been updated to reflect the changes to 3.1 made on 11/6.

adhilto: The change was to clarify the actual requirement. Step 4 was the essential bit. Steps 5 and on, "Optionally...," were and still are optional, and have been removed to avoid confusion.

scubagoggles/rego/Commoncontrols.rego
```diff
@@ -1055,39 +1312,108 @@ tests contains {
 ########################
 
 #
-# Baseline GWS.COMMONCONTROLS.9.1v0.3
+# Baseline GWS.COMMONCONTROLS.9.1
```
adhilto, line 1315: GWS.COMMONCONTROLS.9.1 and 9.2, even with the policy API, are not able to be tested using ScubaGoggles. The baseline requires that "Highly privileged accounts" and "All sensitive user accounts" be enrolled. The best ScubaGoggles could currently do is tell you if users are allowed to enroll, it can't tell you who is actually enrolled. The reports API actually has this capability too, but we removed that from ScubaGoggles so users wouldn't get a false sense of security.

scubagoggles/rego/Commoncontrols.rego
```diff
@@ -1574,6 +1988,12 @@ if {
 # GWS.COMMONCONTROLS.12 #
 #########################
```
adhilto, line 1929: The policy API implementation of GWS.COMMONCONTROLS.12.1 does not check for the takeout status of services without individual admin center control (steps 4 and 5 of the [implementation](https://github.com/cisagov/ScubaGoggles/blob/main/baselines/commoncontrols.md#gwscommoncontrols121v03-instructions)). The corresponding setting in the policy API appears to be `takeout_service_status`.

scubagoggles/rego/Commoncontrols.rego
rlxdev: done

scubagoggles/rego/Commoncontrols.rego
rlxdev: done

docs/installation/DownloadAndInstall.md
```diff
@@ -1,56 +1,205 @@
...
+  C:\Users\userID\scubagoggles
+Setup: OPA executable directory
+(WARNING): OPA executable not found in PATH
+Location of OPA executable [C:\Users\userID\scubagoggles]?
+  downloading: opa_windows_amd64.exe
+  OPA executable: C:\Users\userID\scubagoggles\opa_windows_amd64.exe
+Setup: Google API credentials file
+Google credentials (JSON) file [C:\Users\userID\scubagoggles\credentials.json]?
+(WARNING): Google credentials file not found in C:\Users\userID\scubagoggles\credentials.json
+```
```
snarve, line 182: Couple of things here:

- This includes multiple prompts 
- Each prompt asks if user would like to continue the path provided or change it, if the goal is to have all the files centrally located in the user's home directory then why provide an option to change the path where this can be installed
- This could also lead to the user saving different setup files in different locations (ex: opa in one place and credentials in another), this defeats the purpose of a centrally allocated setup files folder
- The above given sequence does not work, since the steps to setup the credentials is after this, so there is no credentials file available for the user to save yet
- Also, the user will then have to place the credentials in this folder, its easier to have everything in the project folder itself
- In the above setup process opa is already installed, and then in the next steps there is a link to install opa, this needs to be updated
- all the setup files can be installed later after they have been created and downloaded, without the prompts to simplify the process

docs/installation/OPA.md
```diff
@@ -1,43 +1,49 @@
...
-
-## Navigation
-- Continue to [Prerequisites](/docs/prerequisites/Prerequisites.md)
-- Return to [Documentation Home](/README.md)
+
+# Download the OPA executable
+
+The tool makes use of [Open Policy Agent's Rego Policy language](https://www.openpolicyagent.org/docs/latest/policy-language/).
+An OPA executable is required to execute this tool and can be downloaded
+with the `scubagoggles getopa` command.
```
snarve, line 6: Need to switch the sequence since opa is already installed in the setup process

docs/installation/DownloadAndInstall.md
```diff
@@ -1,56 +1,205 @@
...
+  downloading: opa_linux_amd64_static
+  OPA executable: /home/userID/scubagoggles/opa_linux_amd64_static
+Setup: Google API credentials file
+Google credentials (JSON) file [/home/userID/scubagoggles/credentials.json]?
+(WARNING): Google credentials file not found in /home/userID/scubagoggles/credentials.json
+```
+
+## Navigation
+- Continue to [Download the OPA executable](OPA.md)
+- Return to [Documentation Home](/README.md)
```
snarve, line 205: The download and install file has large amount of text which is helpful to understand the process, however from a user's perspective this is confusing (to find instructions within the text) and a lot of information.
The instructions can be simplified with a shorter version with clear steps on what to do next.
The informative text can be saved in separate place for reference.

snarve (changes requested):

scubagoggles/rego/Drive.rego
rlxdev: After implementing Google's policy "reduction" process, this is working for me
when I change settings in the Admin UI.

scubagoggles/rego/Drive.rego
rlxdev: After implementing Google's policy "reduction" process, this is working for me
when I change settings in the Admin UI.

scubagoggles/rego/Drive.rego
rlxdev: After implementing Google's policy "reduction" process, this is working for me
when I change settings in the Admin UI.

scubagoggles/rego/Drive.rego
rlxdev: After implementing Google's policy "reduction" process, this is working for me
when I change settings in the Admin UI.

scubagoggles/rego/Drive.rego
rlxdev: After implementing Google's policy "reduction" process, this is working for me
when I change settings in the Admin UI.

scubagoggles/rego/Drive.rego
rlxdev: After implementing Google's policy "reduction" process, this is working for me
when I change settings in the Admin UI.

scubagoggles/rego/Commoncontrols.rego
rlxdev: After implementing Google's policy "reduction" process, this is working for me
when I change settings in the Admin UI.

scubagoggles/scuba_argument_parser.py
```diff
@@ -1,137 +1,184 @@
...
+        path_value_options = ('credentials',
+                              'documentpath',
+                              'opapath',
+                              'outputpath',
+                              'regopath')
+
+        for option_name in path_value_options:
+            if (option_name in args
+                and not isinstance(getattr(args, option_name), Path)):
+                setattr(args, option_name, path_parser(args.credentials))
```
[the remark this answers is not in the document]
rlxdev: I seem to have lost my mind at some point when I made this change, but it should work correctly now.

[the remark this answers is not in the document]
adhilto: It happens. And I can confirm, it works now.

scubagoggles/rego/Commoncontrols.rego
adhilto: This still isn't right. Say I enable takeout for Blogger, Books, and for services without individual admin control. The details then says, "The following apps with individual admin control have Takeout enabled: Blogger, Google Books." No mention of services without individual admin control. Now say I enable takeout for Blogger and Books, but disable it for services without individual admin control. Now the report goes green, when it should still fail and report that takeout is enabled for Blogger and Books.

adhilto (approved):
Great job addressing my feedback. I still have a few concerns, but I don't see them as big enough to block the policy API, so I propose we open follow-up issues for those and merge this as it is now.

My remaining concerns:
- The Windows version of the smoke test is failing. It passes when run locally and the Mac version passes, so a recommend disabling the Windows workflow for now and revisiting it later.
- The installation process. I am no longer concerned by the so-called "conflicting" config files I brought up earlier, but still share some of David and Sheetal share some of these concerns. Also though I definitely see your changes here as a step in the right direction. I agree with David that there should be a .scubagoggles folder in the home directory, where the .scubagoggles initial config file lives. I also recommend that be the default location for the OPA executable. I also agree with Sheetal that prompts presented by the setup process are confusing and that some of the documentation needs tweaking.
- The issue with Takeout for services without individual admin control described in my other comment.

Again, I'm ok with these final things being addressed later (as long as we open issues). Thanks for addressing all my other feedback!

snarve (commented):
Pending comments addressed in issue: https://github.com/cisagov/ScubaGoggles/issues/531

snarve (approved):
Great job Roy!

Any pending comments are added in issue: https://github.com/cisagov/ScubaGoggles/issues/531

Most of the comments are about the setup process.

docs/installation/DownloadAndInstall.md
```diff
@@ -1,56 +1,205 @@
...
+to display the download site in a browser.  For installing ScubaGoggles as a
+user (and not a developer), you should download the file with the name ending
+in `.whl` (known as a Python "wheel" file), for example
+`scubagoggles-1.0.0-py3-none-any.whl`.
+
+If you are a developer, you may alternatively download the "gzipped tar" file
+(file ending with `.tar.gz`), or zip file (file ending with `.zip`).  You may
+also clone the GitHub repository.  The instructions that follow focus on the
+installation for the general user.
+
```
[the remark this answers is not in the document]
snarve: noted in https://github.com/cisagov/ScubaGoggles/issues/531

docs/installation/DownloadAndInstall.md
```diff
@@ -1,56 +1,205 @@
...
+which ScubaGoggles is installed must be able to access the internet so the
+dependencies may be downloaded.
+
+## Creating the ScubaGoggles Output Directory
+ScubaGoggles produces "Secure Baseline Conformance Reports", which are written
+to a directory on your system.  In addition, ScubaGoggles requires the OPA
+executable and the Google credentials file.  It is recommended that you create
+a directory that will contain these required files and serve as a location
+for the reports.  You will run the ScubaGoggles setup utility to indicate the
+location of the output directory.
```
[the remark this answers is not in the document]
snarve: noted in https://github.com/cisagov/ScubaGoggles/issues/531

docs/installation/DownloadAndInstall.md
snarve: noted in https://github.com/cisagov/ScubaGoggles/issues/531

docs/installation/DownloadAndInstall.md
snarve: noted in https://github.com/cisagov/ScubaGoggles/issues/531

docs/installation/OPA.md
snarve: noted in https://github.com/cisagov/ScubaGoggles/issues/531

PR #672, opened 2025-04-15
Better error message for when authorized scopes do not match

### LINKED ISSUE #635: Better error message for when authorized scopes do not match
# 💡 Summary #
When ScubaGoggles permissions are updated but the user has not updated their credentials, OAuth or Service Account with the new scopes this error message is thrown.
```
('access_denied: Requested client not authorized.', {'error': 'access_denied', 'error_description': 'Requested client not authorized.'})
```

This issue is to make the error message more user friendly by informing the user what OAuth 2.0 scopes they're 
missing. Something similar to `('access_denied: Requested client not authorized.', {'error': 'access_denied', 'error_description': 'Requested client not authorized.'})`
` Your credential is missing the "https://www.googleapis.com/auth/admin.directory.customer.readonly" scope`

Stack trace from @snarve 
Image
## Implementation notes ##

- Updates to likely only `auth.py`. 
- Would it be possible to detect whether a service account or OAuth credential is used and display a targeted error message?

k-winters:
## 🗣 Description ##

Provides a better error message for when the authorized scopes do not match. There didn't seem to be a way to directly verify the authorized scopes for a service account, so if the service account does not have all the authorized scope, the error just indicates that one of the expected scopes are missing. The error message for OAuth, does indicate which scope is missing.

### 💭 Motivation and context
Closes #635 

## 🧪 Testing
For both OAuth and service accounts, change the authorized scopes for the user and verify the error message is shown.

- [X] These code changes follow [cisagov code standards](https://github.com/cisagov/development-guide).
- [X] All new and existing tests pass.

## ✅ Pre-merge Checklist

## ✅ Post-merge Checklist

### DIFF
--- a/scubagoggles/provider.py
+++ b/scubagoggles/provider.py
@@ -693,5 +693,5 @@ class Provider:
         # {'error': 'access_denied', 'error_description': 'Requested client not authorized.'})
         scopes_list = self._credentials.scopes
         if 'access_denied: Requested client not authorized.' in str(exc):
-            warnings.warn(f'Your credential may be missing one' 
+            warnings.warn(f'Your credential may be missing one'
                           f' of the following scopes: {scopes_list}')
--- a/scubagoggles/provider.py
+++ b/scubagoggles/provider.py
@@ -688,9 +688,10 @@ class Provider:
         return results
 
     def _check_scopes(self, exc):
-        # If one of the scopes is not authorized in a Service account the 
-        # error is thrown: ('access_denied: Requested client not authorized.', 
+        # If one of the scopes is not authorized in a Service account the
+        # error is thrown: ('access_denied: Requested client not authorized.',
         # {'error': 'access_denied', 'error_description': 'Requested client not authorized.'})
-        scopes_list = GwsAuth._scopes
+        scopes_list = self._credentials.scopes
         if 'access_denied: Requested client not authorized.' in str(exc):
-            warnings.warn(f'Your credential may be missing one of the following scope: {scopes_list}')
+            warnings.warn(f'Your credential may be missing one' 
+                          f' of the following scopes: {scopes_list}')
--- a/scubagoggles/provider.py
+++ b/scubagoggles/provider.py
@@ -403,6 +403,7 @@ class Provider:
                 f'Exception thrown while getting top level OU: {exc}',
                 RuntimeWarning
             )
+            self._check_scopes(exc)
             self._unsuccessful_calls.add(ApiReference.LIST_OUS.value)
             return 'Error Retrieving'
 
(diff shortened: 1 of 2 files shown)


adhilto (changes requested):
This only partially fixes this issue (at least with service accounts, that's what I tested). Printing a warning doesn't seem like adequate severity. Basically, this adds a new warning but doesn't suppress the other warnings or exception tracebacks that occur as a result. So while the proper cause of the problem is identified, it's lost in a long stream of warnings and exceptions.

Ultimately, shouldn't ScubaGoggles stop once it detects there's a problem? Currently it will keep trying to make API calls even it detects there are missing scopes.

My recommendations:
1. For the error handling in the `get_toplevel_ou` function, add an `except` block for `RefreshError` specifically. While other problems could cause a refresh error, we know that an OAUTH scopes problem will raise that kind of exception. This step isn't strictly necessary, but it's best practice to catch the more specific type of exception when you can.
2. Leave the original `except` block unchanged, and inside the new block for the RefreshError, check the scopes, and change `warnings.warn` to `logging.error` if there's a problem. Additionally, if there's a scopes problem, re-raise the exception so that execution will stop.
3. In main.py, add an `except` block to catch the `RefreshError` specifically (by all the other `except` blocks, at the end of the file.

Doing all that could take that long exception that I can't even fit into a single screenshot to something like this:
image

scubagoggles/provider.py
```diff
@@ -685,3 +690,13 @@ def _get_list(resource, item: str, **list_args) -> list:
...
+
+    def _check_scopes(self, exc):
+        # If one of the scopes is not authorized in a Service account the
+        # error is thrown: ('access_denied: Requested client not authorized.',
+        # {'error': 'access_denied', 'error_description': 'Requested client not authorized.'})
+        scopes_list = self._credentials.scopes
+        if 'access_denied: Requested client not authorized.' in str(exc):
+            logging.error(f'Your credential may be missing one'
+                          f' of the following scopes: {scopes_list}')
+        raise
```
adhilto, line 702: Shouldn't this `raise` be inside the `if`? If it's outside, it will throw the exception even if the error is unrelated to the scopes.

For example, the second check scopes call in the screenshot.
image

If an exception is thrown related to the scopes, it will be caught by the `RefreshError` handler. But if a different kind of exception is thrown, it will be caught by the generic handler. That generic handler calls the check scopes function, then gets interupted early because that function re-raises the exception even though it's not related to the scopes.

```suggestion
            raise
```

adhilto (changes requested):
Thanks for addressing my earlier feedback! Just one more thing.

rlxdev (changes requested):

scubagoggles/provider.py
```diff
@@ -398,11 +399,15 @@ def get_toplevel_ou(self) -> str:
             ou_name = response['name']
             self._successful_calls.add(ApiReference.LIST_OUS.value)
             return ou_name
```
rlxdev, line 403: Pylint is giving this error about the `get_toplevel_ou()` method:

`scubagoggles\provider.py:372:4: R1710: Either all return statements in a function should return an expression, or none of them should. (inconsistent-return-statements)`

Lines 399-401 can easily (and probably should) be moved outside the try/except block and put at the end of the method.  This will get rid of the Pylint warning.

This is caused by this:

```
        except RefreshError as exc:
            self._check_scopes(exc)
```

Pylint isn't "smart enough" to figure out that you're unconditionally raising an exception in `_check_scopes()`, and figures there might be a return path without a value.  Since your method definitely raises an exception, you can move that code I suggested and it'll be correct.

You will also need to add an ending newline at the end of the file, as Pylint is complaining about that, too.

adhilto (approved):

PR #862, opened 2025-12-16
Unit tests for PolicyAPI class

### LINKED ISSUE #776: Create unit tests for policy_api.py
### Prerequisites

### 💡 Summary

Create a new function, test_policy_api.py, which will house the unit tests for the [policy_api.py](https://github.com/cisagov/ScubaGoggles/blob/main/scubagoggles/policy_api.py) function.

The test_policy_api.py function should be stored inside of the scubagoggles/Testing/Python directory.

scubagoggles

* Testing
  * Functional
  * Rego
  * Python
    * test_policy_api.py

### Motivation and context

This issue is in relation to epic #330 

### Implementation notes

We will use pytest library for implementation unit test, use a class-based approach for the file structure.

For the _get method, use a pytest-mock fixture to mock query requests

Class TestPolicyAPI

test_close
test__get_policies
test__get_ou
test__get_groups
test__get_policies_list
test__get
test__reduce
test__merge
test__apply_defaults
test__dump
test_verify

Then running the unit tests are as simple as:

```
# for all Python unit tests
pytest ./Testing/Python/

# for all unit tests for a given file
pytest ./Testing/Python/test_policy_api.py 

# for an individual unit test
pytest ./Testing/Python -k test__get_policies_list
```

### Acceptance criteria

- [ ] The test_policy_api.py function is added
- [ ] TestPolicyAPI class is implemented with methods for each function declared in policy_api.py

rlxdev:
## 🗣 Description ##

This is the implementation of unit tests for the `PolicyAPI` class in the `policy_api.py` module.

### 💭 Motivation and context

Unit tests were developed for the processing of the policy data returned by Google and used in ScubaGoggles, particularly in the Rego code.  Unit test data is stored separately in JSON-formatted files in the data subdirectory below the test code.  The format of the file names used for the test data is: `policy_api_<test-target><subtest-number>.json`, where `<test-target>` indicates the test the data is for (e.g., `get_ou`, `get_policies`, etc.), and `<subtest-number>` is the number of the subtest (starting with 1) to allow for various combinations of test data.

The tests cover the modification and verification of Google’s policy data.  Because its purpose is only to receive response data from Google’s API, the `_get()` method is not covered by a unit test.  Neither is the `_dump()` method, which is only used for debugging/diagnostic purposes.  I don't see a big ROI for spending the time to create a unit test for these, but if the reviewer disagrees please let me know.  My main focus on the `PolicyAPI` class testing are the areas that work with the Policy API data and are susceptible to regression if the code is modified.

Closes #776

## 🧪 Testing

Ran tests using `pytest` on Windows and Ubuntu 22.04 & 24.04, RedHat 8.10.  All tests pass.

- [X] These code changes follow [cisagov code standards](https://github.com/cisagov/development-guide).
- [X] Tests have been added and/or modified to cover the changes in this PR.
- [X] All new and existing tests pass.

## ✅ Pre-merge Checklist

## ✅ Post-merge Checklist

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,8 @@
-# This package is used for software development and is not needed by
+# These packages are used for software development and are not needed by
 # the Scubagoggles users:
 build
+pytest>=9
+pytest-mock
 
 # There are many Python versions for users to install for ScubaGoggles, and
 # because of this, it is too restrictive to limit versions of the dependent
--- a/scubagoggles/Testing/Unit/Python/test_policy_api.py
+++ b/scubagoggles/Testing/Unit/Python/test_policy_api.py
@@ -1,0 +1,559 @@
+"""Tests for the PolicyAPI class.
+"""
+
+import json
+import re
+
+from pathlib import Path
+
+import pytest
+
+# The import is needed as stated and not how pylint wants it.
+# pylint: disable=consider-using-from-import
+
+import google.auth.transport.requests as requests
+import scubagoggles.auth as auth
+
+from scubagoggles.policy_api import PolicyAPI
+
+
+class MockGwsAuth:
+
+    """Mocks the GwsAuth class - the tests in this module do not call any
+    Google API.
+    """
+
+    # pylint: disable=missing-function-docstring
+    # pylint: disable=too-few-public-methods
+
+    @property
+    def credentials(self):
+        return None
+
+
+class MockSession:
+
+    """Mocks Google's AuthorizedSession class.  This along with the above
+    class are used to allow the instantiation of the PolicyAPI class without
+    making any calls to the Google API.
+    """
+
+    # pylint: disable=missing-function-docstring
+
+    def __init__(self, credentials):
+        self._close_count = 0
+        self._credentials = credentials
+
+    def close(self):
+        self._close_count += 1
+
+    def get(self, *_):
+        pass
+
+auth.GwsAuth = MockGwsAuth
+
+requests.AuthorizedSession = MockSession
+
+
+class TestPolicyApi:
+
+    """This class contains unit tests for the PolicyAPI class.
+    """
+
+    # The test module needs to access "internal" methods.
+    # pylint: disable=protected-access
+
+    _data_dir = Path(__file__).parent / 'data'
+
+    def test_close(self, monkeypatch):
+
+        """Tests that the AuthorizedSession is closed when the PolicyAPI is
+        used as a context manager.  Closing the session causes the resources
+        to be cleaned up.
+        """
+
+        self._patch_policy_api(monkeypatch)
+
+        with PolicyAPI(auth.GwsAuth(), 'topOU') as policy_api:
+            session = policy_api._session
+
+        assert session._close_count == 1
+
+    @pytest.mark.parametrize(('value', 'expected'),
+                             (# Valid values
+                              (True, True),
+                              (False, True),
+                              # Invalid values
+                              (1100, False),
+                              ('true', False),
+                              ('False', False)))
+    def test_isbool(self, value, expected):
+
+        """Tests for the isBool() lambda used in the PolicyAPI expected
+        settings.
+        """
+
+        assert PolicyAPI.isBool(value) is expected
+
+    @pytest.mark.parametrize(('value', 'expected'),
+                             (# Valid values
+                              ('TEST', True),
+                              ('TRUSTED_DOMAINS', True),
+                              # Invalid values
+                              (32768, False),
+                              ('192.168.0.1', False)))
+    def test_isenum(self, value, expected):
+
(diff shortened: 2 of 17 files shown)


scubagoggles/Testing/Unit/Python/test_policy_api.py
```diff
@@ -0,0 +1,548 @@
...
+
+        assert session._close_count == 1
+
+    def test_isbool(self):
+
+        """Tests for the isBool() lambda used in the PolicyAPI expected
+        settings.
+        """
+
+        valid_values = (True, False)
```
mitchelbaker-cisa: This seems like a good candidate for [parametrize](https://docs.pytest.org/en/stable/how-to/parametrize.html). We could do something like the following:

```python
@pytest.mark.parametrize(
  ("value", "expected"),
  [
    # Valid values
    (True, True),
    (False, True),
    # Invalid values
    (1100, False),
    ('true', False),
    ('False', False),
  ]
)
def test_isbool(self, value, expected):
  """Tests for the isBool() lambda used in the PolicyAPI expected
  settings.
  """

  result = PolicyAPI.isBool(value)
  assert result == expected
```

Same for the test_isenum, test_isint, etc.

requirements.txt
```diff
@@ -1,6 +1,8 @@
-# This package is used for software development and is not needed by
+# These packages are used for software development and are not needed by
 # the Scubagoggles users:
 build
+pytest>=9
+pytest-mock
```
mitchelbaker-cisa, line 5: We also need to include `pytest-subtests` package in requirements.txt

scubagoggles/Testing/Unit/Python/test_policy_api.py
rlxdev: Hi Mitchel,

I've converted these `test_isXXX()` methods over to using `pytest.mark.parametrize()`.

requirements.txt
rlxdev: I was going to include `pytest-subtests` in `requirements.txt` but then I found that it is **already included** in the `pytest v9+` packages, so that's why instead I put the requirement that `pytest` be at least `v9`.

scubagoggles/Testing/Unit/Python/test_policy_api.py
mitchelbaker-cisa: Looks good, thanks.

requirements.txt
mitchelbaker-cisa: Ah okay, I must have an older version of pytest installed because the tests using subtests failed without the pytest-subtests package installed. I'll remove it and test with v9 of pytest.

mitchelbaker-cisa (approved):
Tested with pytest v9.x, pytest-subtests is included as expected. Everything looks good to me.

dagarwal-ecs (changes requested):
Quick question: Rather than creating these data files which have very similar json files with additional OU or other differences, why don't we do something similar to how we do it on ScubaGear where we have a baseline json configuration, and then json patch into it the sections we want to change, when we run the tests, in order to not create so many large files where we can accidentally miss a typo. Not sure if something like this is possible or not, but wanted to open the discussion.

mitchelbaker-cisa:
> Quick question: Rather than creating these data files which have very similar json files with additional OU or other differences, why don't we do something similar to how we do it on ScubaGear where we have a baseline json configuration, and then json patch into it the sections we want to change, when we run the tests, in order to not create so many large files where we can accidentally miss a typo. Not sure if something like this is possible or not, but wanted to open the discussion.

It's possible. Pytest doesn't have a built-in feature for this but wouldn't be hard to roll our own. We would have to add helper method(s) to import the base JSON file and create a deep copy of it. Then you could update the deep copy as you see fit for each unit test. 

Simplified example:
```python
import copy

# Base JSON
base_json = {
    "users": [
        {"primaryEmail": "admin@example.com"}
    ]
}

# Perform deep copy
test_data = copy.deepcopy(base_json)

# Overwrite existing property
test_data["users"][0]["primaryEmail"] = "newemail@example.com"

# Add a new property to existing object
test_data["users"][0]["isAdmin"] = True

# Add a new user to the list
test_data["users"].append({"primaryEmail": "abc@example.com"})

# Add a new top-level key
test_data["domains"] = ["example.com", "test.org"]
```

Updated JSON structure:
```json
{
    "users": [
        { "primaryEmail": "newemail@example.com", "isAdmin": true },
        { "primaryEmail": "abc@example.com" }
    ],
    "domains": [ "example.com", "test.org" ]
}
```

The above example modifies properties directly which is fine for simple cases, but we also have nested JSON. We'd likely need to write a helper method that merges existing keys with the patched update we're providing, as to not lose existing data from the base JSON.

@rlxdev Thoughts?

PR #875, opened 2026-01-07
Create a CI Pipeline

### LINKED ISSUE #702: Create a CI pipeline
### Prerequisites

### 💡 Summary

We currently run our linting and unit testing workflows separately during pull requests and against new commits added to main. This works fine as is but creating a dedicated CI pipeline will help with scalability as we add functional testing, security checks, etc.

### Motivation and context

Creating a dedicated CI pipeline can be connected to a specific environment, for example `testpypi`. As new commits are added to main, we can run linting/unit tests/publish workflows and test ScubaGoggles in real time. 

### Implementation notes

1. Add `workflow_call:` to [pylint.yml](https://github.com/cisagov/ScubaGoggles/blob/main/.github/workflows/pylint.yml) and [run_opa_tests.yml](https://github.com/cisagov/ScubaGoggles/blob/main/.github/workflows/run_opa_tests.yml) to make them reusable workflows. They can then be called from other workflows.
2. Remove events that trigger pylint.yml/run_opa_tests.yml from main/pull_request/etc since the CI pipeline will handle this
3. Create new run_pipeline.yml which bundles the CI jobs together, reference [run_pipeline.yml](https://github.com/cisagov/ScubaGear/blob/main/.github/workflows/run_pipeline.yaml) from ScubaGear.

### Acceptance criteria

- [ ] Existing workflow triggers are modified such that they can only be run manually (workflow_dispatch) or via a call from another workflow (workflow_call)
- [ ] A CI pipeline workflow can be run against pull requests/main

ObedAmpahECS:
## 🗣 Description ##

Creates a dedicated CI pipeline that helps to consolidate automated checks into a single file, `run_pipeline.yml`, and adds functionality to control rerunning checks (ex, pytest).

### 💭 Motivation and context

Resolves #702 

## 🧪 Testing

1. Pull this branch into your local
2. Rename it
3. Push the branch up and open a new draft PR (run_smoke_test only runs on specific `[pull_request]` event triggers as it is a resource-expensive workflow
4. Confirm passing pipeline checks for 
- Pylint
- OPA Tests
- Markdown links
- Python Tests
- Smoke Test

## ✅ Pre-merge Checklist

- [x] Squash all commits into one PR level commit using the `Squash and merge` button.

## ✅ Post-merge Checklist

- [x] Delete the branch to clean up.
- [x] Close issues resolved by this PR if the [closing keywords](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword) did not activate.

### DIFF
--- a/.github/workflows/run_pytest_coverage.yml
+++ b/.github/workflows/run_pytest_coverage.yml
@@ -25,4 +25,5 @@ jobs:
       - name: Run tests
         run: |
           python -m pip install coverage
-          coverage report source=. -m pytest ./scubagoggles/Testing/Unit/Python -vvv
+          coverage run -m pytest ./scubagoggles/Testing/Unit/Python -vvv
+          coverage report -m


.github/workflows/run_smoke_test.yml
```diff
@@ -6,15 +6,15 @@ on:
...
-      - "*smoketest*"
+  # push:
+  #   # These paths are primarily for workflow testing purposes
+  #   paths:
+  #     - ".github/workflows/run_smoke_test.yml"
+  #     - ".github/actions/setup-dependencies-windows/action.yml"
+  #     - ".github/actions/setup-dependencies-macos/action.yml"
+  #   branches:
+  #     - "main"
+  #     - "*smoketest*"
```
rlxdev:
```suggestion
  push:
    # These paths are primarily for workflow testing purposes
    paths:
      - ".github/workflows/run_smoke_test.yml"
      - ".github/actions/setup-dependencies-windows/action.yml"
      - ".github/actions/setup-dependencies-macos/action.yml"
    branches:
      - "main"
      - "*smoketest*"
```

Was this supposed to be **temporarily** disabled while testing the branch to avoid the Policy API quota failures?

.github/workflows/run_smoke_test.yml
```diff
@@ -180,15 +185,15 @@ jobs:
             /home/runner/scubagoggles
           key: ${{ runner.os }}-baseline-${{ needs.linux-path.outputs.cachepath }}
```
rlxdev:
```suggestion
          key: ${{ runner.os }}-baseline-${{ needs.smoke-test.outputs.linux-path }}
```
This: `needs.linux-path.outputs.cachepath` is actually **undefined**, so the key winds up being `Linux-baseline-` and it still finds the cached folder because the folder starts with that.  This change just adds the correct completed path.

.github/workflows/run_smoke_test.yml
```diff
@@ -159,6 +159,11 @@ jobs:
           echo $FOUNDLINUX
           echo "cachepath=$FOUNDLINUX" >> $GITHUB_OUTPUT
 
+      - name: Vulnerability Scan
+        uses: pypa/gh-action-pip-audit@v1.1.0
+        with:
+          inputs: requirements.txt
+
```
rlxdev, line 167:
```suggestion
        continue-on-error: true
```
I had thought that adding this vulnerability scan was going to be OK, but I just ran it and it's complaining about `pygments`, which is a dependency of `pytest`.  It's complaining about version 2.19.2 which happens to be the latest version.  **If there's a vulnerability, this step fails and the workflow stops**.  I added this to make this step optional because otherwise we'll never get a complete run until pygments puts out a new release.

Unfortunately, with this `continue-on-error`, you get a "green" and it continues, so you're not likely to notice a potential vulnerability unless you look at the run log.  The other potential issue is that this doesn't only look at our direct dependencies, but at the indirect dependencies (like pygments).  It doesn't apply in this case, but sometimes a package requires a particular version of a dependency and if it's got a vulnerability there's nothing we can do about the sub-dependency (pygments) if we need the dependency (pytest).

.github/workflows/run_smoke_test.yml
ObedAmpahECS: Yes, thanks for catching that 🙏

.github/workflows/run_smoke_test.yml
ObedAmpahECS: Hmm, ok I can look at other vulnerability scan packages or move to a manual test, taking a look at this now

ObedAmpahECS: Screenshot 2026-03-30 at 7 10 22 AM

I just sent a run through with `no-deps: true` to tell pip audit to do a direct audit of the requirements.txt listed dependency (avoids sub-dependencies like pyments). Allows for passing tests, but could be viewed as an incomplete scan against all possible vulnerabilities.

.github/workflows/pylint.yml
```diff
@@ -34,4 +22,4 @@ jobs:
         pip install pylint pytest selenium
     - name: Analysing the code with pylint
       run: |
-        pylint -d R0913,R0914,R0915,R1702,W0718,W0719,R0801 $(git ls-files '*.py')
+        pylint -d R0913,R0914,R0915,R1702,W0718,W0719,R0801 $(git ls-files '*.py')
```
mitchelbaker-cisa, line 25: Thoughts on revisiting why these pylint rules are disabled? 

R0913: too-many-arguments
R0914:	too-many-locals
R0915:	too-many-statements
R1702:	too-many-nested-blocks
W0718: broad-exception-caught
W0719: broad-exception-raised
R0801:	duplicate-code

I think things like too-many-statements, too-many-nested-blocks, etc. cause unnecessary noise, but R0801 on catching duplicate code seems useful. I ran `pytest $(git ls-files '*.py')` and it indicated the MockSession class is declared twice.

Image

.github/workflows/run_pytest_coverage.yml
```diff
@@ -0,0 +1,29 @@
...
+        with:
+          python-version: 3.14
+
+      - name: Install dependencies
+        run: |
+          python -m pip install -r requirements.txt
+
+      - name: Run tests
+        run: |
+          python -m pip install pytest-mock coverage
```
mitchelbaker-cisa: pytest-mock is already installed in the previous step with `python -m pip install -r requirements.txt`. This seems like it can be removed, unless the dependencies don't carry over from one step to another?

Image

.github/workflows/run_pytest_coverage.yml
```diff
@@ -0,0 +1,29 @@
+name: Pytest
+on:
+  workflow_call:
+  workflow_dispatch:
+
+
+permissions:
+  contents: read
+  pull-requests: read
```
mitchelbaker-cisa: I presume `contents: read` is required by actions/checkout to clone the repo. Does this workflow require `pull-request: read` for anything? If not recommend we remove.

.github/workflows/run_pipeline.yml
```diff
@@ -0,0 +1,22 @@
+name: Run CI Pipeline for ScubaGoggles
+
+on:
+  push:
+  workflow_dispatch:
+
+
+permissions: read-all
```
mitchelbaker-cisa: `permissions: read-all` grants read access to all scopes: contents, pull_requests, deployments, etc. If we don't require these scopes then recommend we downgrade to whatever the least permissions required, otherwise the `read-all` scope will be inherited by all the sub jobs. 

```
permissions:
  contents: read
```

.github/workflows/run_smoke_test.yml
rlxdev: I actually prefer this (no deps).  We can control our own dependencies, provided there's a new version to cover the one identified with the vulnerability.  With the sub-dependencies, I can see this check failing and stopping the workflow, and we don't have direct control over these - and there are quite a few of them.

.github/workflows/pylint.yml
rlxdev: I actually know about the duplicate because it's my code.  Even with the pylint error, I'm not sure I'd want to do anything about it because these are small "fake" classes just intended for testing (I have thought about seeing if there's an alternative).  For this case, I'd wind up disabling it anyway.   FWIW, I also think some of their limits are a bit arbitrary (like maximum 4 arguments in a function).

.github/workflows/run_pytest_coverage.yml
ObedAmpahECS: This was left-over from a package test, will remove!

.github/workflows/run_pipeline.yml
ObedAmpahECS: Testing and running, will confirm and make necessary edit, thanks.

.github/workflows/run_pytest_coverage.yml
ObedAmpahECS: That was the issue prior but moving to a manual run of coverage made this unnecessary, thanks!

rlxdev (approved):
This looks good to me.  As a reminder, once this has been merged into main, the pytest runs will fail until my PR #976 is also merged into main because it has fixes for the policy API unit tests.

PR #954, opened 2026-03-16
780 allow user to specify doh servers

### LINKED ISSUE #780: Allow user to specify preferred DoH servers
### Prerequisites

### 💡 Summary

Allow user to specify preferred DoH servers, similar to how specifying preferred DNS resolvers (see #645).

### Motivation and context

Some users may be unable to resolve SPF, DKIM, or DMARC using their network's internal resolver due to split horizon issues, use a public DNS resolver due to firewall restrictions, or use Cloudflare DoH due to firewall restrictions. Specifying custom DoH servers would be the user's only path forward in this case.

### Implementation notes

This is unfortunately more complicated than simply replacing https://github.com/cisagov/ScubaGoggles/blob/3a4e2e96fe1c82c848d533ce2aaada8004df0fc3/scubagoggles/robust_dns.py#L127 with a user-provided list.
- The code currently assumes that the HTTP endpoint is `dns-query`, but this actually isn't mandated by the RFC.
- The code currently actually doesn't use DoH as defined by [RFC 8484](https://www.rfc-editor.org/rfc/rfc8484). It instead just uses a JSON API provided by Cloudflare. Simpler to implement, but not universal.

Fortunately, DNS python (the library ScubaGoggles uses), has implemented a DNS over HTTPS method. See https://dnspython.readthedocs.io/en/stable/query.html#dns.query.https. We'd need to switch to using that rather than the `requests` library for the DoH queries.

Recommend we add a `preferreddohservers` CLI argument that accepts a list of strings that would replace the hard-coded list defined on the line linked above.

### Acceptance criteria

- [ ] We use true DoH rather than the simple Cloudflare JSON API
- [ ] We add a CLI argument that allows users to specified their preferred DoH servers

pvanderveen110:
## 🗣 Description ##

This feature change allows users to specify preferred Domain Name Servers over HTTPS (DoH) when DoH queries are called, instead of using the pre-defined list of DoH servers that was originally hard coded. This feature also replaces the API call to requests with one to a true DoH request using the dns module instead.

### 💭 Motivation and context

Closes #780 

## 🧪 Testing

I manually isolated and tested the logic required to replace the current functionality of retrieving text with request.get to dns.message.make_query and dns.query.https. I verified the same text results and number of results are returned using both the original logic (requests) and the DoH request methods and attributes. 

## ✅ Pre-merge Checklist

## ✅ Post-merge Checklist


adhilto (changes requested):
Smoke tests are failing. I also get the same error when I run ScubaGoggles manually:
```
    self.resolver = dns.resolver.Resolver()
                    ^^^^^^^^^^^^
AttributeError: module 'dns' has no attribute 'resolver'
```

Unrelated, but Parameters.md and full_config.yaml also still need to be updated.

docs/usage/Parameters.md
```diff
@@ -84,6 +85,11 @@ options:
                         IP addresses of DNS resolvers that should be used to retrieve any DNS records required by
                         specific SCuBA policies. Optional; if not provided, the system default will be used. Example:
                         --preferreddnsresolvers 8.8.8.8 8.8.4.4
+  --preferreddohservers <doh-servers>
+                        IP addresses of DoH servers that should be used to retrieve any TXT records (i.e. required 
+                        by specific SCuBA policy). Optional; If not provided, the DoH query will use a default 
+                        list of DoH servers: "cloudflare-dns.com", "[2606:4700:4700::1111]", "1.1.1.1". Example:
+                        --preferreddohservers 8.8.8.8 cloudfare-dns.com
```
adhilto: 1. Instead of "IP addresses of DoH servers..." say "IP addresses or domain names of DoH servers..."
2. Clarify that the DoH queries will only be made if the traditional DNS queries fail. Maybe something like "...used to retrieve any TXT records, should the traditional DNS queries fail..."
3. Take off the brackets surrounding "[2606:4700:4700::1111]". Leaving the brackets in would probably make the user think they needed to input IPv6 addresses with the brackets, but the brackets aren't actually a part of the address so I think users would find that counterintuitive.

scubagoggles/sample-config-files/full_config.yaml
```diff
@@ -94,6 +94,11 @@ outputreportfilename: BaselineReports
 #     - 8.8.8.8
 #     - 8.8.4.4
 
+# Use this field to control what DNS over HTTPS (DoH) servers are used (if specified) execute DoH queries.
+# preferreddohservers:
+#     - 8.8.8.8
+#     - cloudflare-dns.com
+
```
adhilto: Two things:
- Indent with 2 spaces, not 4. Technically either is fine for YAML, but we need to be consistent. Our sample YAML is already inconsistent (there's an issue open for fixing that, but most of the examples in it use 2 spaces).
- Give an IPv6 example too

```suggestion
#   - 8.8.8.8
#   - cloudflare-dns.com
#   - 2606:4700:4700::1111
```

scubagoggles/robust_dns.py
```diff
@@ -139,14 +145,23 @@ def get_doh_server(self) -> str:
...
+
+                # Attempt to resolve a.root-servers.net over DoH if no preferred list is specified.
+                # The domain chosen is somewhat arbitrary, as we don't care what the answer is,
+                # only if the query succeeds/fails. a.root-servers.net, the address of one of the
+                # DNS root servers, was chosen as a benign, highly-available domain.
+                if self.preferred_doh_list is None:
+                    uri = f"https://{server}/dns-query?name=a.root-servers.net"
+
                 headers = {"accept":"application/dns-json"}
                 requests.get(uri, headers=headers, timeout=2).json()
```
adhilto: The get_doh_server function also needs to be updated to use the dns library rather than requests

scubagoggles/robust_dns.py
```diff
@@ -190,32 +205,43 @@ def doh_query(self, qname : str, max_tries : int) -> dict:
         try_number = 0
         while try_number < max_tries:
             try_number += 1
-            uri = f"https://{self.doh_server}/dns-query?name={qname}&type=txt"
-            headers = {"accept":"application/dns-json"}
+
+            # perferred server is already in URI format
```
adhilto: This doesn't actually appear to be the case, at least if I'm following the code correctly. `self.doh_server` is set to whatever `self.get_doh_server()` returns, and that function will return either none or an item from the doh_servers` list, so it'll just be a domain name or IP address.

scubagoggles/robust_dns.py
```diff
@@ -190,32 +205,43 @@ def doh_query(self, qname : str, max_tries : int) -> dict:
...
+            uri = self.doh_server
+
+            # Defeault DoH server is not a URI
+            if self.preferred_doh_list is None:
+                uri = f"https://{self.doh_server}/dns-query?name={qname}&type=txt"
+
+            #headers = {"accept":"application/dns-json"}
```
adhilto: Remove OBE code
```suggestion
```

scubagoggles/robust_dns.py
```diff
@@ -190,32 +205,43 @@ def doh_query(self, qname : str, max_tries : int) -> dict:
...
             try:
-                response = requests.get(uri, headers=headers, timeout=5).json()
-                if response['Status'] == 0:
+
+                #response = requests.get(uri, headers=headers, timeout=5).json()
+                # True DoH
```
adhilto: Remove commented out code
```suggestion
```

scubagoggles/robust_dns.py
```diff
@@ -190,32 +205,43 @@ def doh_query(self, qname : str, max_tries : int) -> dict:
...
             try_number += 1
-            uri = f"https://{self.doh_server}/dns-query?name={qname}&type=txt"
-            headers = {"accept":"application/dns-json"}
+
+            # perferred server is already in URI format
+            uri = self.doh_server
+
+            # Defeault DoH server is not a URI
+            if self.preferred_doh_list is None:
+                uri = f"https://{self.doh_server}/dns-query?name={qname}&type=txt"
```
adhilto: A few things here: 
1. There needs to be some validation of `self.doh_server`. Specifically, if the user inputs an IPv6 address, the IPv6 address needs to be wrapped in square brackets to form a valid URL, and I don't think it's fair to expect users to handle that.
2. Change to `f"https://{self.doh_server}/dns-query`. The `neme=` and `type=` parts are not actually part of the formal DoH protocol, those were just needed for the Cloudflare API.
3. There is a potential issue with hardcoding the path `/dns-query`. That path isn't actually mandated by the [RFC](https://datatracker.ietf.org/doc/html/rfc8484). The RFC uses that as an example, and all the DoH providers I've looked at use that path, but technically DoH providers don't have to use that path. We could just hard-code the path like you have here since it's an edge case (I'm not sure in practice how often people change it) or a more robust solution would be to create another ScubaGoggles parameter, something like `dohpath`, and have that default to `/dns-query`.

scubagoggles/main.py
```diff
@@ -220,6 +220,16 @@ def gws_dispatch(args):
                         metavar='<dns-resolvers>',
                         help=help_msg)
 
+    help_msg = ('IP addresses of DoH servers that should be used to retrieve any'
+                'TXT records (i.e. required by specific SCuBA policy). Optional;'
+                'If not provided, the DoH query will use a default list of DoH servers:\n'
+                '[\"cloudflare-dns.com\", \"[2606:4700:4700::1111]\", \"1.1.1.1\"]')
```
adhilto: The comment I added for Parameters.md applies here too

docs/usage/Parameters.md
```diff
@@ -84,6 +85,11 @@ options:
                         IP addresses of DNS resolvers that should be used to retrieve any DNS records required by
                         specific SCuBA policies. Optional; if not provided, the system default will be used. Example:
                         --preferreddnsresolvers 8.8.8.8 8.8.4.4
+  --preferreddohservers <doh-servers>
```
ahuynhECS: added this to my PR #965 and added you as a reviewer. Maybe suggest removing due to potential merge conflicts.

adhilto (changes requested):

docs/usage/Parameters.md
ahuynhECS: added your suggestions @alden & @pvanderveen110 to my PR #965 https://github.com/cisagov/ScubaGoggles/blob/858-revamp-parametersmd/docs/usage/Parameters.md#preferred-doh-servers

docs/usage/Parameters.md
ahuynhECS: https://github.com/cisagov/ScubaGoggles/blob/858-revamp-parametersmd/docs/usage/Parameters.md#preferred-doh-servers

scubagoggles/main.py
pvanderveen110: Can you please clarify on this? You said Parameters.md needs to be updated and it is with the exact same help message as main.py.   So what do I need to update here?

pvanderveen110: Oh wait, I see now. You wanted me to edit Parameters.md and update this accordingly too. Right? Ah my bad.

adhilto: Yep! Though you can hold off on Parameters.md now since Andrew updated that, but still need to update the message in main.py.

scubagoggles/robust_dns.py
pvanderveen110: I was thinking of retaining the original DoH server list and just hardcoding the "https://{server}/dns-query?name=a.root-servers.net" URL for each "server" in the original list.

pvanderveen110: I'll certainly add in a fix where we can detect if the address is IPv6 and just add the square brackets around it.

ahuynhECS (approved):
no additional comments from me aside from resolving the merge conflict. You can remove the parameter.md changes since they are covered in PR #965

scubagoggles/robust_dns.py
```diff
@@ -138,26 +145,41 @@ def get_doh_server(self) -> str:
...
+                if re.match(pattern, server):
+                    server = "[" + server + "]"
+
+                uri = f"https://{server}/dns-query"
+
+                # Attempt to resolve DoH server if no preferred list is specified.
+                # The domain chosen is somewhat arbitrary, as we don't care what the answer is,
+                # only if the query succeeds/fails.
+
+                query = dns.message.make_query(uri, dns.rdatatype.TXT)
```
adhilto: The first param to the make_query is the qname, that is to say, the name you're trying to resolve. So it should be a domain name, not the URL path to the DoH server. You can use the qname the code was using before.
```suggestion
                query = dns.message.make_query("a.root-servers.net", dns.rdatatype.TXT)
```

scubagoggles/robust_dns.py
```diff
@@ -186,36 +208,47 @@ def doh_query(self, qname : str, max_tries : int) -> dict:
...
+
         # DoH is available, query for the domain
         try_number = 0
         while try_number < max_tries:
             try_number += 1
-            uri = f"https://{self.doh_server}/dns-query?name={qname}&type=txt"
-            headers = {"accept":"application/dns-json"}
+
+            # form the DoH query
+            qname = f"https://{self.doh_server}/{dohpath}"
```
adhilto: So this is problematic. `qname` is a function parameter. You're overwriting it here without ever reading what the value was. You're going to want to do something similar to what you're doing in the get_doh server function:
```suggestion
            uri = f"https://{self.doh_server}/{dohpath}"
```

scubagoggles/robust_dns.py
```diff
@@ -186,36 +208,47 @@ def doh_query(self, qname : str, max_tries : int) -> dict:
...
+
             try:
-                response = requests.get(uri, headers=headers, timeout=5).json()
-                if response['Status'] == 0:
+
+                #response = requests.get(uri, headers=headers, timeout=5).json()
+                # True DoH
+                query = dns.message.make_query(qname, dns.rdatatype.TXT)
+                response = dns.query.https(query, qname, timeout=5)
```
adhilto: The change pairs with the above suggested change
```suggestion
                response = dns.query.https(query, uri, timeout=5)
```

adhilto (changes requested):
This is looking much better!

Some more comments:
1. Need to fix the merge conflicts that Andrew pointed out.
2. Need to make Pylint happy, see the latest run on this branch here: https://github.com/cisagov/ScubaGoggles/actions/runs/23828116872/job/69455561991?pr=954
3. It actually looks like we need to update the ScubaGoggles dependencies for this. I executed this in a venv setup just for ScubaGoggles, and I got this error when I tried to make a DoH query using DNS python: `dns.query.NoDOH: DNS-over-HTTP3 is not available.`. Running `pip3 install dnspython[doh]` fixed this.
4. See below for additional changes needed

pyproject.toml
adhilto, line 1: Don't forget to update https://github.com/cisagov/ScubaGoggles/blob/main/requirements.txt as well

scubagoggles/robust_dns.py
adhilto: Quick reminder about this comment

pyproject.toml
pvanderveen110: Done

adhilto (approved):
Thanks for addressing all my feedback!

Issue #950, opened 2026-03-05
Notify users if a new release is available


adhilto:
### Prerequisites

- [x] This issue has an informative and human-readable title.

### 💡 Summary

When users run ScubaGoggles, check to see if a new release is available and notify the users.

### Motivation and context

So users are using the latest version. Will be especially important after the 1.0 release.

### Implementation notes
Example showing how to get the installed version of ScubaGoggles:
```
from scubagoggles.version import Version
print(Version.number)
```

Example showing how to query what the latest version is from PyPi: https://stackoverflow.com/questions/28774852/pypi-api-how-to-get-stable-package-version.

I recommend performing this check whenever the user runs `scubagoggles gws` and printing a warning if a new version is available.

### Acceptance criteria

Users are informed if a new release is available when they run ScubaGoggles.

adhilto:
> [@adhilto](https://github.com/adhilto) will refine this issue.

@FollyBeachGurl I added more details.

PR #971, opened 2026-03-30
Notify user if a new version is available

ehazard-mitre:
## 🗣 Description ##

This will check if the local ScubaGoggles version is behind the latest version available on PyPI and will suggest to the user to update to the newest version if they are behind. This check occurs after argument parsing occurs and will not print anything to the user if they are on the latest version. 

### 💭 Motivation and context

Want to ensure the user is running the latest version of ScubaGoggles. This closes Issue #950. 

## 🧪 Testing

Tested by manually assigning version numbers (e.g., local is 0.1.0 when latest available is 0.5.0) to test for logic performance. Also included error testing if the HTTP request to the PyPI JSON file is unavailable and tested with incorrect URLs for bad HTTP requests. 

## ✅ Pre-merge Checklist

## ✅ Post-merge Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -20,7 +20,8 @@ dependencies = ["dnspython>=2.6.1",
                 "MarkupSafe>=2.1.5",
                 "pyyaml>=6.0.2",
                 "requests>=2.32.3",
-                "tqdm>=4.66.5"]
+                "tqdm>=4.66.5",
+                "packaging>=26.0"]
 requires-python = ">= 3.10"
 description='SCuBA security baseline assessment tool'
 authors = [{name = "CISA"}]


scubagoggles/orchestrator.py
```diff
@@ -136,6 +140,35 @@ def gws_products(cls) -> dict:
...
+        try:
+            goggles_url_response = requests.get(SCUBAGOGGLES_PACKAGE_URL, timeout=10)
+            goggles_url_response.raise_for_status()
+            latest_version_on_pypi = goggles_url_response.json()["info"]["version"]
+        except Exception as err:
+            error_message = (
+                f"An unexpected error occurred in retrieving latest SCuBA Goggles"
+                f"version: {err}"
+            )
+            print(error_message)
```
adhilto: Use log.error instead of print here

scubagoggles/orchestrator.py
```diff
@@ -136,6 +140,35 @@ def gws_products(cls) -> dict:
...
+            return
+
+        if local_machine_version < latest_version_on_pypi:
+            local_goggles_install_is_behind = (
+                f"A new version of SCuBA Goggles is available (v{latest_version_on_pypi}),"
+                f" you're running {local_machine_version}.\nConsider updating via: "
+                f"pip install scubagoggles."
+            )
+            print(local_goggles_install_is_behind)
```
adhilto: Use log.warn instead of print here

scubagoggles/orchestrator.py
```diff
@@ -136,6 +140,35 @@ def gws_products(cls) -> dict:
...
+            latest_version_on_pypi = goggles_url_response.json()["info"]["version"]
+        except Exception as err:
+            error_message = (
+                f"An unexpected error occurred in retrieving latest SCuBA Goggles"
+                f"version: {err}"
+            )
+            print(error_message)
+            return
+
+        if local_machine_version < latest_version_on_pypi:
```
adhilto, line 168: We'll need something more robust than a string compare. For example, string compare would say that 1.9 is greater that 1.11.

scubagoggles/orchestrator.py
```diff
@@ -136,6 +140,35 @@ def gws_products(cls) -> dict:
...
+            local_goggles_install_is_behind = (
+                f"A new version of SCuBA Goggles is available (v{latest_version_on_pypi}),"
+                f" you're running {local_machine_version}.\nConsider updating via: "
+                f"pip install scubagoggles."
```
adhilto: Need to specify upgrade otherwise pip won't do anything since an older version of ScubaGoggles is already installed.
```suggestion
                f"pip install scubagoggles --upgrade."
```

scubagoggles/orchestrator.py
ehazard-mitre: Could we use `packaging.version`? This requires installing `packaging` (which adds another requirement in `requirement.txt`). More info here: https://stackoverflow.com/questions/11887762/how-do-i-compare-version-numbers-in-python.

scubagoggles/orchestrator.py
ehazard-mitre: Thank you for reviewing Alden!

scubagoggles/orchestrator.py
adhilto: New dependencies are ok. Just be sure to update  https://github.com/cisagov/ScubaGoggles/blob/main/requirements.txt.

scubagoggles/orchestrator.py
```diff
@@ -136,6 +141,38 @@ def gws_products(cls) -> dict:
...
+            log.error(error_message)
+            return
+
+        local_machine_version = packageVersion.Version(local_machine_version)
+        latest_version_on_pypi = packageVersion.Version(latest_version_on_pypi)
+
+        if local_machine_version < latest_version_on_pypi:
+            local_goggles_install_is_behind = (
+                f"A new version of SCuBA Goggles is available (v{latest_version_on_pypi}),"
+                f" you're running {local_machine_version}.\nConsider updating via: "
```
atuomit: Small change add "v" in front of local_machine_version.

atuomit (changes requested):
Small change, suggest adding "v" in front of local_machine_version.

scubagoggles/orchestrator.py
atuomit: @adhilto Since requirements.txt was modified does the requires_dist portion of https://pypi.org/pypi/scubagoggles/json need to be updated? How would we do that?

adhilto (approved):

scubagoggles/orchestrator.py
adhilto: I'm actually not sure. @mitchelbaker-cisa, do you know?

ehazard-mitre: I don't know for certain but I think PyPI takes requirements from the `pyproject.toml`file and will update the `requires_dist` metadata (e.g., it'll update `requires_dist` in the next version that's uploaded to PyPI using the new `pyproject.toml` file). I just pushed another commit to add packaging to the pyproject file.

mitchelbaker-cisa: @adhilto @ehazard-mitre Yep pypi refers to the pyproject.toml file when building the release bundle, should be good here.

PR #1048, opened 2026-06-02
ci: restrict run_link_checker_cron_job workflow permissions to read-only

arpitjain099:
Adds an explicit least-privilege `permissions:` block to `.github/workflows/run_link_checker_cron_job.yml`. The job only reads the repository, so `contents: read` is sufficient; today the workflow inherits whatever the repository default grants, which is broader than it needs. This is one of the checks OpenSSF Scorecard flags under [Token-Permissions](https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions).


billbirk:
Noticed this issue extends to a few other workflows — opened #1050 to cover `pylint.yml`, `run_link_checker_for_md_edits.yml`, and `run_opa_tests.yml` in case it's helpful.

arpitjain099:
> Noticed this issue extends to a few other workflows — opened #1050 to cover `pylint.yml`, `run_link_checker_for_md_edits.yml`, and `run_opa_tests.yml` in case it's helpful.

Thank you @billbirk

arpitjain099:
Thanks for catching that and picking up the rest, @billbirk. You're right that the same read-only scoping applies to those (pylint, the md-edits link checker, and the opa tests all just read the repo). Between this PR and #1050 the workflow set is covered. I'm happy to defer to whatever the maintainers prefer here, whether that is taking both or consolidating into one - either works for me.

dagarwal-ecs:
Closing this PR since there are no files changed

PR #1058, opened 2026-06-10
Generate OSCAL catalog release artifact for SCuBA baselines

### LINKED ISSUE #1004: Generate a machine-readable version of the SCBs
### Prerequisites

### 💡 Summary

The purpose of this issue is to produce machine-readable JSON formats of the SCuBA markdown SCBs. The JSON data should be available in the `scubagoggles/` project directory and as a release asset. The first step for this issue should be creating a script that converts markdown files to JSON format.

After that, the next important consideration is our team frequently drafts new baseline updates. We want end-to-end automation of the JSON SCB generation during the PR process when new baseline updates are introduced, to remove manual intervention required to update the JSON SCBs and improve future maintainability. Below is an example of how this might work:

1. The team creates new baseline updates and create a new PR (normal operation with updated markdown)
2. Trigger a GitHub action based on a `pull_request` event that builds the JSON SCB files. Iterate through the files changed in the PR and compare them with a set, something like { drivedocs.md, gemini.md, calendar.md, etc. }, that includes valid markdown file names, or under the `pull_request` trigger simply list out all of the markdown SCB file names. Exit the action early if the PR did not make any changes to these file names, otherwise run the action.
3. github-actions bot makes a new commit to the PR with the equivalent JSON SCB updates.
4. The new baseline update PR will now include both the markdown and JSON SCBs. 

As for publishing as a release asset, assume we have a file like `scubagoggles/schema/scuba_baselines.json` in the project after the script to convert markdown to JSON and the GitHub action is complete. To publish this file as a release asset, the existing [run_release.yaml](https://github.com/cisagov/ScubaGoggles/blob/main/.github/workflows/run_release.yml) workflow's softprops/action-gh-release job will need to be updated to include the new JSON SCB file. 

### Motivation and context

Similar ScubaGear [issue  ](https://github.com/cisagov/ScubaGear/issues/1886) and [PR](https://github.com/cisagov/ScubaGear/pull/2079)

### Implementation notes

1. First create a script that converts markdown to JSON 
2. Connect this with the new GitHub action, and run the script when a PR is opened that includes baseline updates
3. Update the existing [run_release.yaml](https://github.com/cisagov/ScubaGoggles/blob/main/.github/workflows/run_release.yml) workflow's softprops/action-gh-release job to include the new JSON SCB file

### Acceptance criteria

- [ ] A machine-readable JSON version of the SCuBA SCBs are created
- [ ] A Github action is able to apply new baseline updates after a PR is opened and apply the same respective machine-readable SCB updates to the same PR
- [ ] The machine-readable JSON output file is included as a release asset when run_release.yaml is run

ejaronne:
## Summary

This PR adds generation of a single OSCAL catalog release artifact for the Google Workspace SCuBA baseline Markdown files.

The Markdown baselines remain the source of truth. The generated OSCAL catalog represents the README-listed baselines in their existing hierarchy: baseline groups, section groups, and policy controls. The intent is to provide a machine-readable OSCAL artifact that preserves enough Markdown-derived context for downstream tools to render or process the baselines without using the Markdown documents as the only copy of the content.

This supersedes the earlier component-definition-oriented approach in this PR. The current implementation does not generate one component definition per NIST mapping and does not reorder the source content around NIST SP 800-53 controls. Instead, it keeps the SCuBA baseline structure intact and records NIST mappings as properties on the relevant SCuBA policy controls.

Closes #1004.

## Changes

- Adds `scubagoggles/utils/generate_oscal_baselines.py` to generate one OSCAL catalog from the baseline Markdown files listed in `scubagoggles/baselines/README.md`.
- Adds a `Generate OSCAL Baselines` workflow to produce the `oscal-baseline-catalog` artifact.
- Adds OSCAL catalog generation to the release workflow so release artifacts include the generated catalog.
- Uses `DEFAULT_OSCAL_VERSION = "1.2.2"` for the generated catalog metadata and filename.
- Validates the generated catalog with pinned OSCAL CLI `3.2.0`.
- Documents the Markdown-to-OSCAL mapping and modeling assumptions in [OSCAL catalog documentation](docs/misc/oscal-catalog.md).

## Testing/Validation

- Generate OSCAL Baseline Catalog workflow passed with no annotations/warnings and produced the `oscal-baseline-catalog` artifact:
  https://github.com/ejaronne/ScubaGoggles/actions/runs/31241261093/artifacts/9017076026
- Local OSCAL CLI validation passed for the generated catalog.
- OSCAL generator unit tests passed: `3 passed`.
- Full Python unit test suite passed: `567 passed, 32 subtests passed`.
- Workflow YAML parse check passed.
- Current PR smoke-test matrix is failing in the fork/PR context; I have left "All new and existing tests pass" unchecked until a maintainer can run or confirm the credentialed smoke tests.

- [X] Tests have been added and/or modified to cover the changes in this PR.
Note: I have not checked "All new and existing tests pass" because the Run Smoke Test workflow appears to depend on repository secrets unavailable to forked PRs.

## ✅ Pre-merge Checklist

## ✅ Post-merge Checklist


.github/workflows/generate_oscal_baselines.yml
```diff
@@ -0,0 +1,52 @@
+name: Generate OSCAL Baselines
+
+on:
+  pull_request:
```
mitchelbaker-cisa: Can we remove the `pull_request` event triggers? This was causing issues on the ScubaGear side which caused the remote branch to become out of sync with the changes on a developer's local branch. 

I'm fine with the `push` events, this would run the workflow when commits are merged into main, allowing us to see an updated version of the OSCAL baselines. If the files listed under `push.paths` are modified in a future issue, then the workflow would also be triggered against a feature branch as well which is helpful for development purposes. `workflow_dispatch` is important to keep as it allows us to run the workflow manually as needed.

scubagoggles/utils/generate_oscal_baselines.py
```diff
@@ -0,0 +1,917 @@
...
+    generated_at = oscal_timestamp()
+    doc = {
+        "component-definition": {
+            "uuid": stable_uuid(namespace, "component-definition"),
+            "metadata": {
+                "title": f"Google Workspace {area} SCuBA Baseline - OSCAL Component Definition",
+                "published": generated_at,
+                "last-modified": generated_at,
+                "version": version,
+                "oscal-version": "1.1.2",
```
mitchelbaker-cisa: I'd prefer the OSCAL version to not be hardcoded here. Maybe this is abstracted out into its own helper function that calls the GitHub API to pull the latest version from the usnistgov/OSCAL releases page? 

Also, how do we want to handle maintenance as they release new versions? I'm thinking of the scenario where they update the OSCAL schema and we'd need to reflect that change in our machine-readable SCBs.

scubagoggles/utils/generate_oscal_baselines.py
```diff
@@ -0,0 +1,917 @@
...
+    return doc, summary
+
+
+def parse_section(
+    text: str,
+    section_matches: list[re.Match[str]],
+    section_idx: int,
+    source_policy_count: int,
+    covered_areas: list[str],
+    implemented: list[dict[str, Any]],
```
mitchelbaker-cisa: Is there a more specific type that can be used here instead of `Any` for `implemented` and `resources` parameters?

mitchelbaker-cisa (changes requested):
Great work Eugene, as a first pass I added a few comments below for your review. I plan on doing a second pass to take a closer look at the JSON files stored in the release asset, [oscal-baselines.zip](https://github.com/user-attachments/files/28988087/oscal-baselines.zip).

.github/workflows/generate_oscal_baselines.yml
ejaronne: fixed with: #[bf2f8d2](https://github.com/cisagov/ScubaGoggles/pull/1058/commits/bf2f8d2b1bc8051c50753b686150809dc4204250)

ejaronne:
For discussion purposes, I'm attaching a Calendar component definition:
[calendar-oscal-component-definition.json](https://github.com/user-attachments/files/29108952/calendar-oscal-component-definition.json)

ejaronne:
Since first introduction, we've been having internal design discussions, possibly changing or adding to the types of formats from the different OSCAL layers and approaches. Discussions continue, including reaching out to OSCAL SMEs.

scubagoggles/utils/generate_oscal_baselines.py
ejaronne: With the help of codex, here's an approach I reviewed with it:

Good point. I’m planning to treat the OSCAL version in the generator as the validated/supported OSCAL version rather than automatically tracking the latest NIST release during generation.

I can move this into a `DEFAULT_OSCAL_VERSION` constant and add an `--oscal-version` CLI override so release automation can set it intentionally. For longer-term maintenance, I think a separate scheduled workflow or follow-up issue would be safer: check the latest `usnistgov/OSCAL` release, then open an issue/PR when it differs so maintainers can validate schema compatibility before bumping the constant.

Thoughts?

mitchelbaker-cisa: I'm good with the first approach to add a `DEFAULT_OSCAL_VERSION` constant for now. [scuba_constants.py](https://github.com/cisagov/ScubaGoggles/blob/main/scubagoggles/scuba_constants.py) is where we define version constants.

Creating a follow-up issue to investigate how we want to handle new releases published to usnistgov/OSCAL sounds like a plan for future maintenance. My initial thought is the schema compatibility seems complex. We'd have to compare the before/after of how the OSCAL schema looks, then update generate_oscal_baselines.py accordingly. Doing this for every new OSCAL version would add overhead for our team to manage, but if OSCAL schema updates are less frequent, i.e. included as major versions, then that seems reasonable for us to validate against.

ejaronne:
7/16/2026 meeting with NIST OSCAL team:
- they recommend using the [OSCAL Control Layers:](https://pages.nist.gov/OSCAL/learn/concepts/layer/) Catalog, Profile, and Control Mapping
- our initial artifacts for this PR were implementation (component definition) and assessment (assessment plan, results) layer
- To help users understand how the control layer can support the upper layers, we can still provide snippet examples of the way a component definition, assessment plan, and assessment results would render to OSCAL.
- They also noted that we should update the the latest OSCAL version (1.1.2 -> 1.2.2)

ejaronne:
Update since July 16, 2026:

This PR has been reworked from the earlier component-definition-oriented approach to the current OSCAL catalog approach discussed in review.

The current implementation now generates a single OSCAL catalog for the README-listed Google Workspace SCuBA baseline Markdown files. The Markdown baselines remain the source of truth, and the generated catalog preserves the existing SCuBA hierarchy: baseline groups, section groups, and policy controls.

Additional documentation was added here:

- [OSCAL catalog documentation](https://github.com/ejaronne/ScubaGoggles/blob/feature/oscal-baseline-release-artifacts/docs/misc/oscal-catalog.md)
- [Markdown-to-OSCAL cross-walk and modeling notes](https://github.com/ejaronne/ScubaGoggles/blob/feature/oscal-baseline-release-artifacts/docs/misc/oscal-catalog.md#markdown-to-oscal-cross-walk)
- [OSCAL part-name mapping assumptions](https://github.com/ejaronne/ScubaGoggles/blob/feature/oscal-baseline-release-artifacts/docs/misc/oscal-catalog.md#part-name-mapping)
- [Mappings documentation update](https://github.com/ejaronne/ScubaGoggles/blob/feature/oscal-baseline-release-artifacts/docs/misc/mappings.md)

The generated catalog is validated with a pinned OSCAL CLI version in CI.

A sample generated artifact from this PR branch is available here:

https://github.com/ejaronne/ScubaGoggles/actions/runs/31241261093/artifacts/9017076026

scubagoggles/utils/generate_oscal_baselines.py
ejaronne: Addressed. `DEFAULT_OSCAL_VERSION` is now defined centrally in `scubagoggles/scuba_constants.py` and imported by the OSCAL generator. The generated catalog uses that value for `metadata.oscal-version`, the OSCAL JSON reference link, and the generated filename.

Evidence:
- `DEFAULT_OSCAL_VERSION = '1.2.2'` in `scubagoggles/scuba_constants.py`
- `generate_oscal_baselines.py` imports and uses `DEFAULT_OSCAL_VERSION`
- CI checks that the generated catalog metadata and filename use the pinned version
- Unit tests assert the generated filename and `metadata.oscal-version` match the constant

scubagoggles/utils/generate_oscal_baselines.py
ejaronne: Addressed. The older component-definition path that had the `implemented` parameter has been removed as part of the catalog-only approach. For the remaining generated JSON structures, the generator now defines recursive `JsonValue` and `JsonObject` type aliases and uses those for resource lists, return values, and JSON helper functions instead of `Any`.

Evidence:
- `JsonValue` / `JsonObject` are defined in `scubagoggles/utils/generate_oscal_baselines.py`
- `resources` parameters are now typed as `list[JsonObject]`
- `rg Any scubagoggles/utils/generate_oscal_baselines.py` has no remaining matches

adhilto (commented):
Hi @ejaronne, sorry for the delay. First off, really great work here. I just have a few minor comments:

1. Do we want to reference viewer.oscal.io in the documentation as a tool for users to view the OSCAL catalog? It definitely helped me as I reviewed the PR to visualize the catalog produced. Here is an example of what that tool can produce:

Image

2. For the version in the metadata, I'm not sure if "scuba-google-workspace-baselines" is the right thing to put. If I'm reading the code right, it looks like it'll go with the release version if that is defined, but fallback to that string if not. I'm wondering if we'd just want to put the ScubaGoggles version instead? Is the thought that we want to distinguish the catalog produced when not part of a release? What if we put something like "1.0 (development)" as the version number for ad-hoc runs built outside the release process?

3. Despite the resolve_href function, I'm still seeing several relative links in the prose. Not sure why the function didn't catch them.

sylvesterkaczmarek (commented):
The OSCAL CLI version is pinned, but both workflows execute the downloaded ZIP without verifying its checksum or signature. That makes CI and the release job trust whatever bytes are served at that URL. Verify a pinned digest/signature before adding the CLI to `PATH`, and fail the workflow on a mismatch.

ejaronne:
Posting this here to address @adhilto’s review [comments ](https://github.com/cisagov/ScubaGoggles/pull/1058#pullrequestreview-4974768881)above.

Thanks @adhilto, these have been addressed in the latest updates.

1. Added a “Viewing the Catalog” section to `docs/misc/oscal-catalog.md` that points users to `viewer.oscal.io` as an optional way to browse the generated OSCAL catalog. The docs still distinguish this from validation, which is performed by the pinned OSCAL CLI in CI.

2. Updated ad hoc/local catalog generation so `metadata.version` uses the ScubaGoggles package version with a development build label when `--release-version` is not supplied. Release artifacts still use the release version passed by the release workflow.

3. Updated Markdown prose cleanup so relative Markdown links copied into OSCAL `prose` are resolved to stable absolute GitHub URLs.

Validation performed:
- OSCAL generator unit tests pass locally: `5 passed`.
- Local release-style catalog generation succeeded.
- Local OSCAL CLI `3.2.0` validation passed.
- New workflow artifact: https://github.com/ejaronne/ScubaGoggles/actions/runs/33658205143/artifacts/9857594323

ejaronne:
Thanks @sylvesterkaczmarek, good catch. I addressed this by pinning the SHA-256 digest for the OSCAL CLI 3.2.0 ZIP in both workflows and verifying the downloaded ZIP with `sha256sum --check --strict` before unzipping it or adding the CLI to `PATH`.

I also verified locally that the pinned digest matches the downloaded `oscal-cli-enhanced-3.2.0-oscal-cli.zip` artifact.

.github/workflows/generate_oscal_baselines.yml
```diff
@@ -0,0 +1,78 @@
...
+      - name: Set up Python
+        uses: actions/setup-python@v6
+        with:
+          python-version: "3.12"
+      - name: Set up Java
+        uses: actions/setup-java@v5
+        with:
+          distribution: temurin
+          java-version: "17"
+      - name: Install OSCAL CLI
```
mitchelbaker-cisa, line 33: I noticed the "Setup Java" and "Install OSCAL CLI" steps are identical in both generate_oscal_baselines.yml and in run_release.yml. Can we instead move these steps into its own composite action so its more reusable? There's some examples of how we use composite actions in [run_smoke_test.yml](https://github.com/cisagov/ScubaGoggles/blob/main/.github/workflows/run_smoke_test.yml):
- [setup-dependencies-windows/action.yml](https://github.com/cisagov/ScubaGoggles/blob/main/.github/actions/setup-dependencies-windows/action.yml)
- [setup-dependencies-macos/action.yml](https://github.com/cisagov/ScubaGoggles/blob/main/.github/actions/setup-dependencies-macos/action.yml)

What a composite action also resolves is maintaining the OSCAL CLI version and SHA256 in one location, currently we'd have to remember to update both locations.

.github/workflows/generate_oscal_baselines.yml
```diff
@@ -0,0 +1,78 @@
+name: Generate OSCAL Baseline Catalog
+
+on:
+  push:
+    paths:
+      - "scubagoggles/baselines/**/*.md"
+      - "scubagoggles/scuba_constants.py"
```
mitchelbaker-cisa, line 7: If someone modifies scuba_constants.py and commits their changes to GitHub, then this workflow will be triggered which I don't think we want.

`"scubagoggles/baselines/**/*.md"` makes sense to keep here for testing purposes to compare markdown updates with its OSCAL output, same with the `"scubagoggles/utils/generate_oscal_baselines.py"` and `".github/workflows/generate_oscal_baselines.yml"` files.

adhilto (approved):
Great work!

Note that there are now merge conflicts, but other than that you have a green light from me.

.github/workflows/generate_oscal_baselines.yml
```diff
@@ -0,0 +1,78 @@
...
+          unzip -q "$OSCAL_CLI_ZIP" -d "$RUNNER_TEMP/oscal-cli"
+          echo "$RUNNER_TEMP/oscal-cli/bin" >> "$GITHUB_PATH"
+      - name: Generate OSCAL baseline catalog
+        run: |
+          python3 scubagoggles/utils/generate_oscal_baselines.py \
+            --input-dir scubagoggles/baselines \
+            --output-dir dist/oscal-baseline-catalog
+      - name: Validate generated JSON parses
+        run: |
+          python3 - <<'PY'
```
mitchelbaker-cisa, line 49: Since this python code is run inside of a workflow, pylint won't report any issues and we're unable to unit test it with pytest. Can we move the validation logic into a validate_oscal_baselines.py file or something similar so that we can unit test it?

This validation logic would also ensure the OSCAL catalog generated during development (generate_oscal_baselines.yml) is the same as what's include as a release asset (run_release.yml).

mitchelbaker-cisa (changes requested):
I imported the OSCAL catalog into https://viewer.oscal.io and it looks great, seeing all of the products and their respective policies displayed. I left a few comments below on how we can improve the workflow logic for your review.

sylvesterkaczmarek (commented):
Re-reviewed the downloaded OSCAL CLI verification. Both workflows pin the 3.2.0 SHA-256 and run `sha256sum --check --strict` before unzipping or adding the CLI to PATH. My previous blocker is resolved.

pjavan:
Exciting to see CISA begin to adopt OSCAL for machine readable.

The catalog approach recommended by NIST is currently being discussed among 2 other paths.  The discussion revolves around how benchmarks or hardening guidance should be introduced into OSCAL so that an implementor can claim their posture against the guidance and an assessor can independently assess and capture the result.  

This creates a question on whether "rules" (aka guidance) should be introduced in a Catalog or an implemented requirement in a CDefs.  It also raises the question how 3rd party auditors can leverage these rules in assessment activities.  

The analysis is available here: https://oscal-foundation.github.io/Pattern-Library/docs/analysis/2026-08-automating-technical-hardening-guidance/

You will see CISA BOD 25-01 hardening guides for M365 were one of the 3 hardening guides used to evaluate the approaches. That along with a CIS and STIG hardening guide.

I mention this not to throw alarm or prevent your team from merging this PR and a catalog approach.  It looks great, however there are consequences with each approach that are still being deliberated.

Example: With Scuba Goggles [mapping to 800-53](https://github.com/cisagov/ScubaGoggles/blob/2dda8c88fd021e782be93d317af5086f59cc508a/scubagoggles/mappings/scuba-to-nist-sp-800-53-r5-fedramp-high.csv#L4), the catalog approach requires additional mapping model support (newer model) and creates greater dependency on profile resolution, a more complex OSCAL concept.

Lastly, we would welcome anyone interested in these use cases to participate in conversations going on in the [OSCAL Foundation](https://oscalfoundation.org/).  The foundation is a great venue where implementors and operators move away from conceptual and address implementation.  We even have a technical focus group on this topic [tfg-automated-assessments](https://oscalfoundationlists.org/g/tfg-automated-assessments)

Edited: Fixed analysis URL

.github/workflows/generate_oscal_baselines.yml
ejaronne: Hi @mitchelbaker-cisa , but `DEFAULT_OSCAL_VERSION` lives in scuba_constants.py currently and changes to that value affect the generated catalog filename, metadata, and OSCAL reference links. Perhaps move it to something new like oscal_constants.py?

.github/workflows/generate_oscal_baselines.yml
ejaronne: Addressed in commit 6402491. I moved the shared Java setup and checksum-verified OSCAL CLI installation into the reusable composite action at .github/actions/setup-oscal-cli/action.yml. Both the catalog-generation and release workflows now call that action, so the OSCAL CLI version and SHA-256 checksum are maintained in one location.

.github/workflows/generate_oscal_baselines.yml
mitchelbaker-cisa: @ejaronne That's fair. I'd rather keep it in scuba_constants.py than creating a new oscal_constants.py. The purpose of scuba_constants.py is to have all variables in one location.

The workflow will still trigger if changes are made to scuba_constants.py, even if these changes are unrelated to the default OSCAL version, but it's inexpensive and these extra runs aren't a big issue.

.github/workflows/generate_oscal_baselines.yml
ejaronne: Addressed in commit `e31d080`. I moved the inline validation logic into `scubagoggles/utils/validate_oscal_baselines.py` and added unit tests in `scubagoggles/Testing/Unit/Python/test_validate_oscal_baselines.py`. Both `generate_oscal_baselines.yml` and `run_release.yml` now use the same validator to check the generated filename and catalog metadata before passing the resulting catalog to `oscal-cli validate`. The focused generator and validator tests pass (`8 passed`).

sylvesterkaczmarek (approved):
Rechecked current e31d080. The OSCAL CLI checksum verification remains centralized in the composite action, and the latest change moves catalog validation into a tested Python module shared by both generation and release workflows. My earlier supply-chain concern remains resolved; no blocker from me on this head.

ejaronne:
Updated this branch with the latest cisagov/ScubaGoggles:main in merge commit fd6450a and resolved conflicts in the release workflow, release documentation, and shared constants.
The resolution preserves upstream’s redesigned release-note workflow while integrating OSCAL catalog generation, validation, ZIP creation, and release attachment. It also retains upstream’s OPA version update and the OSCAL version constant.
Validation completed:
- All 587 Python unit tests passed.
- Pylint passed with a 10.00/10 rating.
- Development and release-style catalogs generated successfully from all 11 baselines.
- Catalog metadata validation passed.
- OSCAL CLI schema validation passed for both catalogs.
- Release workflow YAML, OSCAL step ordering, and artifact filename integration were verified.
- OPA strict validation and all 398 Rego tests passed.
The effective PR scope remains unchanged at 14 files, with no unrelated upstream changes included in the final PR diff.

mitchelbaker-cisa (approved):
Thanks for addressing my comments, @ejaronne. Good to go.

PR #1072, opened 2026-06-16
Adds new licensing table

### LINKED ISSUE #81: Collect License Information
Hi Team,

See the below from [Grant via the M365 Github repo](https://github.com/cisagov/ScubaGear/issues/466). Just sending along so we can get ahead of this one in terms of whether or not Goggles could even do this.

dagarwal-ecs:
## 🗣 Description ##

Closes #81
### 💭 Motivation and context
Adds a licensing table with the common controls report with licensing information for the tenant

## 🧪 Testing

In order to test do the followings:
Enable the API (OAuth and service account):
Open Google Cloud Console
Select the same project as your ScubaGoggles credentials file (check the project_id in your JSON)
Go to APIs & Services → Library
Search for Enterprise License Manager API
Click Enable
Wait at least 30 seconds, then run ScubaGoggles

Run common controls report and check the licensing information table
It should match https://admin.google.com/ac/billing/subscriptions

## ✅ Pre-merge Checklist

## ✅ Post-merge Checklist


scubagoggles/provider.py
```diff
@@ -904,6 +921,96 @@ def get_group_settings(self) -> dict:
             self._unsuccessful_calls.add(ApiReference.GET_GROUP.value)
             return {'group_settings': []}
 
+    def get_license_data(self) -> dict:  # pylint: disable=too-many-branches
```
mitchelbaker-cisa: Can we add python unit tests for `get_license_data`?

scubagoggles/reporter/reporter_html.py
```diff
@@ -315,6 +319,39 @@ def build_individual_report_html(*,
...
+    rows = [
+        {
+            'Product Name': sub['product_name'],
+            'Status': sub['status'],
+            'Assigned Licenses': sub['assigned'],
+        }
+        for sub in license_data
+    ]
+
+    html_table = create_html_table(rows)
```
mitchelbaker-cisa, line 354: Can we update the smoke test utils to check if the common controls HTML includes the tenant licensing table?

mitchelbaker-cisa (changes requested):

docs/prerequisites/Prerequisites.md
```diff
@@ -16,7 +16,8 @@ https://www.googleapis.com/auth/admin.directory.rolemanagement.readonly,
 https://www.googleapis.com/auth/admin.directory.customer.readonly,
 https://www.googleapis.com/auth/apps.groups.settings,
 https://www.googleapis.com/auth/cloud-identity.policies.readonly,
-https://www.googleapis.com/auth/cloud-identity.inboundsso.readonly
+https://www.googleapis.com/auth/cloud-identity.inboundsso.readonly,
+https://www.googleapis.com/auth/apps.licensing
```
mitchelbaker-cisa, line 19: Since the `https://www.googleapis.com/auth/apps.licensing` API scope provides access to create/update/delete operations, we'll need to update the ServiceAccount.md documentation to specify how to create a custom read-only license management role to restrict a service account's access to read-only operations. This will need to be a follow-up issue addressed after #1081 is merged, which adds DASA support for service account authentication.

I created a custom role which can be found under `Account -> Admin Roles -> License Management - Read` in our test tenant if you want to see how its setup.

mitchelbaker-cisa (changes requested):

scubagoggles/reporter/reporter_html.py
```diff
@@ -315,6 +319,39 @@ def build_individual_report_html(*,
...
+
+    :param list license_data: list of subscription dicts, each containing
+        product_name, status, and assigned.  None or empty list → returns ''.
+
+    :return: HTML fragment with the subscriptions table, or empty string.
+    :rtype: str
+    """
+
+    if not license_data:
+        return ''
```
mitchelbaker-cisa: Returning an empty string here doesn't provide the user with any indication about whether an error occurred, if no licenses were returned from the API, etc. Can we add some simple text here to capture this better? For example:

"No licenses found." or "An error occurred when collecting license information."

scubagoggles/provider.py
```diff
@@ -904,6 +921,96 @@ def get_group_settings(self) -> dict:
...
+                        if info['count'] > 0:
+                            subscriptions.append({
+                                'product_name': info['name'],
+                                'sku_id': sku_id,
+                                'product_id': product_id,
+                                'status': 'Active',
+                                'assigned': info['count'],
+                            })
+                except Exception:
+                    pass
```
mitchelbaker-cisa: This inner exception handler is what causes the `403 Forbidden: accessNotConfigured (License Data)` error to repeat 10 times. I'd prefer we add some additional logic to provide more context on which product ids are failing. Below is some prototype code:

```python
# Added under imports at the top of provider.py
from googleapiclient.discovery import build
+from googleapiclient.errors import HttpError

...

def get_license_data(self) -> dict:  # pylint: disable=too-many-branches
    api_call = ApiReference.LIST_LICENSE_ASSIGNMENTS.value
    subscriptions = []

    primary_domain = None
    ...

+   # googleapiclient logs every non-2xx response as a WARNING via the
+   # standard logging module regardless of how we handle the error, so
+   # silence it here since we already report failures via our own
+   # consolidated warning below.
+   http_logger = logging.getLogger('googleapiclient.http')
+   previous_log_level = http_logger.level
+   http_logger.setLevel(logging.ERROR)
    try:
        licensing_service = build('licensing', 'v1',
                                  cache_discovery=False,
                                  credentials=self._credentials)

        any_success = False
+       failed_products: dict[str, list[str]] = {}
        for product_id in KNOWN_PRODUCT_IDS:
            try:
                ...  # unchanged: pagination loop + sku_counts aggregation

                any_success = True
                for sku_id, info in sku_counts.items():
                    ...  # unchanged: append to subscriptions
-           except Exception:
-               pass
+           except Exception as exc:
+               cause = f'HTTP {exc.status_code}: {exc.reason}' \
+                   if isinstance(exc, HttpError) else str(exc)
+               failed_products.setdefault(cause, []).append(product_id)

+       if failed_products:
+           causes = '; '.join(
+               f'{cause} (products: {", ".join(products)})'
+               for cause, products in failed_products.items()
+           )
+           warnings.warn(
+               f'Exception thrown while getting license data for '
+               f'{sum(len(p) for p in failed_products.values())} '
+               f'product(s); these will be omitted from the '
+               f'subscription table: {causes}',
+               RuntimeWarning
+           )

        if any_success:
            self._successful_calls.add(api_call)
        else:
            self._unsuccessful_calls.add(api_call)

    except Exception as exc:
        warnings.warn(
            f'Exception thrown while getting license data; '
            f'subscription table will be omitted: {exc}',
            RuntimeWarning
        )
        self._unsuccessful_calls.add(api_call)
+   finally:
+       http_logger.setLevel(previous_log_level)

    return {'license_data': subscriptions}

```

This will output the following:

```
C:\Users\redacted\Documents\Github\ScubaGoggles\scubagoggles\provider.py:1016: RuntimeWarning: Exception thrown while getting license data for 10 product(s); these will be omitted from the subscription table: HTTP 403: Enterprise License Manager API has not been used in project 766863809156 before or it is disabled. Enable it by visiting https://console.developers.google.com/apis/api/licensing.googleapis.com/overview?project=766863809156 then retry. If you enabled this API recently, wait a few minutes for the action to propagate to our systems and retry. (products: Google-Apps, 101047, 101001, 101005, 101039, Google-Vault, 101033, 101040, 101038, 101035)
  warnings.warn(
```

Then I don't think we'll need the additional section in Troubleshooting.md

mitchelbaker-cisa (changes requested):
After adding the license API I was able to see the license data in the common controls report. I have a couple more comments below:

scubagoggles/reporter/reporter_html.py
dagarwal-ecs: Updated with the following message based on case:
API failure (LIST_LICENSE_ASSIGNMENTS in unsuccessful_calls): `"An error occurred when collecting license information."`
API succeeded but no licenses returned: `"No licenses found."`
Also found a bug where license section would show up no matter the product so also updated that.

scubagoggles/reporter/reporter_html.py
dagarwal-ecs: Updated

scubagoggles/provider.py
dagarwal-ecs: Added python unit tests

mitchelbaker-cisa: Ran locally, looks good

scubagoggles/provider.py
dagarwal-ecs: Updated using your sample code as a model, now it looks like this when it is ran without the license API enabled, and added a fail-fast case so if there is an error with the API it doesn't try to still run all 10 products. Also updated the progress bar to provide some more feedback as to how the run is going

mitchelbaker-cisa (approved):
All comments resolved, thanks!

docs/prerequisites/Prerequisites.md
dagarwal-ecs: @mitchelbaker-cisa I added `f'{BASE_AUTH_URL}/apps.licensing':                          AuthFlow.OAUTH | AuthFlow.DASA,` into the prerequisites, but I have left the implementation of the DASA for you @adhilto since it has some errors upstream with Connect

aor7000 (approved):

PR #1095, opened 2026-06-30
Skip disabled Gmail spam override allow lists when scanning for domains

arpitjain099:
A spam override allow list that the admin has disabled still gets parsed for domains, so if it contains a domain it sets warningDomainsFound (or senderDomainsFound) and the org unit gets flagged against GWS.GMAIL.18.1/18.2 even though that list is off. The cause is that the per-list skip in _gmail_domain_addr checks `if not enable`, but at that point `enable` is the attribute name string pulled from settings_list, which is always truthy, so the skip never runs. I changed it to look the attribute up on the override with `override.get(enable)`, which matches what the method docstring says should happen.

I added two unit tests in test_gmail_rules_parser.py covering a disabled list (no marker emitted) and an enabled list with the same domain (marker still emitted, so real findings are not lost). The full Python unit suite passes locally.

### DIFF
--- a/scubagoggles/Testing/Unit/Python/parsers/test_gmail_rules_parser.py
+++ b/scubagoggles/Testing/Unit/Python/parsers/test_gmail_rules_parser.py
@@ -6,7 +6,28 @@ from pathlib import Path
 from scubagoggles.parsers.gmail_rules_parser import GmailRulesParser
 from scubagoggles.Testing.Unit.Python.test_policy_api import CommonMethods
 
-# pylint: disable=too-few-public-methods
+
+def _spam_override_section(enabled: bool):
+
+    """Builds a minimal spam override section with one "hide warning banner"
+    allow list that contains a domain.  The enable flag is set per the
+    argument so both the disabled and enabled cases can be exercised.
+    """
+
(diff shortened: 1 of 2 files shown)


FollyBeachGurl:
@rlxdev will reference this issue in a new Issue/PR.

arpitjain099:
Hi @FollyBeachGurl, just checking in on this one.

Happy either way, I only want to avoid it sitting here half-open. If @rlxdev is picking the fix up on your side, I'm glad to close this and let that supersede it. If it's useful as-is, it's still current and I can rebase whenever you want it.

For reference, the actual bug is small: the per-list skip in `_gmail_domain_addr` does `if not enable`, but at that point `enable` holds the attribute *name* from `settings_list` rather than its value, so it's always truthy and the skip never fires. The result is that a disabled allow list still contributes domains and the org unit gets flagged against GWS.GMAIL.18.1/18.2.

rlxdev:
Thanks for the contribution! I incorporated your changes (along with testing modifications) in PR #1111 , which is being reviewed. I'm closing this PR since the changes will be integrated when the other PR is approved.

PR #1179, opened 2026-08-26
DMARC edge case with organizational domains

### LINKED ISSUE #1059: DMARC edge case with organizational domains
### Prerequisites

### 💡 Summary

Update how ScubaGoggles checks for the organizational domain

### Motivation and context

DMARC records can be stored either under the full domain name (e.g., test365.cisa.dhs.gov), or under the "organizational domain," which is basically the second-level domain plus the tld (e.g., dhs.gov). Our code currently handles everything I've described up to this point.

The edge case that we aren't handling is that technically there exist "tlds" that in fact consist of multiple labels, mostly because of country-code tlds, e.g., fed.us (see [here](https://publicsuffix.org/list/public_suffix_list.dat) for a list maintained by Mozilla). So if the domain we're trying to find the DMARC record for were subdomain.example.fed.us, we would need to check for the DMARC record under two places:
- subdomain.example.fed.us
- example.fed.us

Our code currently would check for:
- subdomain.example.fed.us
- fed.us

Up until very recently, the only real way to handle cases like this was to maintain a hard-coded list of public suffixes. But a new RFC was just released that has a better way to do that is actually achievable ScubaGoggles:: https://www.rfc-editor.org/rfc/rfc9989.html#section-4.10.

### Implementation notes

Follow the algorithm described here: https://www.rfc-editor.org/rfc/rfc9989.html#section-4.10.

### Acceptance criteria

- [ ] ScubaGoggles follows the updated RFC when looking for the DMARC domains: https://www.rfc-editor.org/rfc/rfc9989.html#section-4.10

adhilto:
## 🗣 Description ##
Update our DMARC checks to follow the algorithm described in the latest DMARC RFC, https://www.rfc-editor.org/rfc/rfc9989.html#section-4.10.

### 💭 Motivation and context
Closes #1059.

## 🧪 Testing
- Tested against our tenant, worked as expected.
- Added hardcoded list of real-world domain names (almost 300). Ran with ScubaGoggles on main and on this branch and made sure any differences in the results for those domain names was expected. Reviewers can DM me if they would like this list.

## ✅ Pre-merge Checklist

## ✅ Post-merge Checklist


scubagoggles/provider.py
```diff
@@ -441,6 +441,122 @@ def get_dkim_records(self, domains: set) -> list:
...
+            return {
+                'domain': domain,
+                'rdata': result['answers'],
+                'log': log_entries
+            }
+
+        # The domain does not exist at the author level. Time to "walk" the DNS tree per RFC 9989
+        # section 4.10.
+        labels = domain.split('.')
+        # If there are 8 or more labels, skip to the final 7 labels. Otherwise, just remove 1 label
```
pvanderveen110, line 511: Note to self: start reviewing here, pick up where we left off.

pvanderveen110 (approved):

sylvesterkaczmarek (commented):
`_parse_dmarc_record()` rejects valid RFC 9989 syntax because it requires the literal `v=DMARC1;` prefix and never trims whitespace around `=`. The ABNF permits WSP around both `=` and `;`, so `v = DMARC1; p = reject` is valid but gets discarded, which can make the tree walk choose a different parent/PSD policy. Parse/strip tags according to the ABNF and add whitespace variants to the DNS cases.

adhilto:
@sylvesterkaczmarek Good catch. I've addressed your feedback in my latest commit.

sylvesterkaczmarek (approved):
The DMARC parser now normalizes RFC-permitted horizontal whitespace around `=` and `;` before checking the required `v=DMARC1` first tag, and the parameterized cases cover the whitespace forms. This resolves the valid-record rejection I raised.

scubagoggles/provider.py
```diff
@@ -441,6 +442,132 @@ def get_dkim_records(self, domains: set) -> list:
...
+
+        # Per the RFC spec, the equals sign and semicolon deliminators can be surounded by 0 or more
+        # space or horizontal tab characters on either side (see
+        # https://www.rfc-editor.org/info/rfc9989/#section-4.8). Strip out those characters to
+        # ensure consistent parsing.
+        answers = [re.sub(r"[ \t]*=[ \t]*", "=", a) for a in answers if a]
+        answers = [re.sub(r"[ \t]*;[ \t]*", ";", a) for a in answers if a]
+
+        # DMARC records MUST start with the "v" tag, with "DMARC1" being the only acceptable value
+        answers = [a for a in answers if a.startswith("v=DMARC1;")]
```
cristina479: RFC 9989 Section 4.8 permits a valid DMARC record containing only v=DMARC1. The semicolon and subsequent tags are optional, and p defaults to `none` when absent. As written, the tree walk can ignore a valid DMARC record and select a less-specific parent record.

Suggest accepting both v=DMARC1 and v=DMARC1;..., and add a regression case in [scubagoggles/Testing/Unit/Python/provider/dns_cases.py](https://github.com/cisagov/ScubaGoggles/blob/1059-dmarc-edge-case-with-organizational-domains/scubagoggles/Testing/Unit/Python/provider/dns_cases.py) (used by [test_provider.py](https://github.com/cisagov/ScubaGoggles/blob/1059-dmarc-edge-case-with-organizational-domains/scubagoggles/Testing/Unit/Python/provider/test_provider.py)).

scubagoggles/Testing/Unit/Python/provider/dns_cases.py
cristina479, line 328: Add unit-test cases under `GET_DMARC_RECORDS_CASES` for the new RFC 9989 tree-walk branches in scubagoggles/provider.py.

For example:
- `psd=n` at provider.py:526: when an ancestor publishes `v=DMARC1; psd=n`, the lookup must stop and return that record as the organizational-domain policy.
- `psd=y` at provider.py:533: when an ancestor publishes `v=DMARC1; psd=y`, the lookup must stop. If the domain immediately below it has a valid DMARC record, return that record; otherwise, return the PSD record.
- Fallback selection at provider.py:553: when no record has `psd=n` or `psd=y`, query through the root and return the valid record with the fewest DNS labels (the closest valid ancestor to the root).
- Query-limit behavior at provider.py:510-515: use an author domain with more than eight labels and assert that the first ancestor lookup begins at the final seven labels, so the author lookup plus the walk cannot exceed eight queries.

cristina479 (changes requested):

adhilto:
@cristina479 Thanks for your feedback. I've made some changes, ready for another review.

sylvesterkaczmarek (approved):
Rechecked current `b549e3fc`. The DMARC parser still normalizes RFC-permitted horizontal whitespace around `=` and `;`, accepts the semicolon-optional `v=DMARC1` form, and the added cases cover the RFC 9989 tree-walk branches. All current CI checks are green. No remaining blocker from my review.

scubagoggles/Testing/Unit/Python/provider/dns_cases.py
cristina479: The fallback case at provider.py (line 463) was not tested. Add a case with multiple valid ancestor records, none containing psd=n or psd=y, and assert that the record nearest the root is returned.

scubagoggles/Testing/Unit/Python/provider/test_provider.py
cristina479, line 326: `query_side_effect()` silently returns an empty response for any unexpected DNS query. Therefore, the long-domain test in `dns_cases.py:687` can pass even if the implementation queries labels that RFC 9989 requires it to skip for domains longer than eight labels because those extra queries produce no log entries and the test only asserts a minimum query count.

Suggest asserting the exact DNS query sequence for the long-domain case, or make unexpected query names fail the test case.

cristina479 (approved):

cristina479 (changes requested):
Approved this review by mistake. Correcting my earlier approval; please address the remaining test coverage feedback.

cristina479 (approved):

cristina479 (approved):

cristina479 (changes requested):
Approved this review by mistake. Correcting my earlier approval; please address the remaining test coverage feedback.

scubagoggles/Testing/Unit/Python/provider/dns_cases.py
cristina479: The fallback case when no record has psd=n or psd=y was not addressed.

scubagoggles/Testing/Unit/Python/provider/test_provider.py
adhilto: I might need more help understanding the issue here, because if I understand correctly, the test case already does this.

So the test doesn't just assert a minimum query count, it also asserts equality, `assert sorted_result == sorted_expected`. The test case does spell out the expected query sequence, e.g., `_dmarc.a.b.c.d.e.f.g.h.i.j.mail.example.com`, `_dmarc.g.h.i.j.mail.example.com`, etc. If I modify the test case so that the expected log uses a different sequence, the test case fails, even if the query count is the same. For example, if I change the second expected query to be `_dmarc.f.g.h.i.j.mail.example.com` (currently line 742 of dns_cases.py), the test fails because the domains don't match.
image

Let me know if I'm missing something.

adhilto: Ok, I think I figured it out and just pushed an update so that unexpected queries get logged now.

sylvesterkaczmarek (approved):
Rechecked current `a731014f`. The only change since my last approval makes unexpected DMARC DNS lookups visible to the result comparison, so the long-domain regression can no longer pass by silently swallowing an extra query. Production code is unchanged and the current CI suite is green. No remaining blocker from me.

sylvesterkaczmarek (approved):
Rechecked current fcc6053d. The latest change strengthens the fallback-domain case by adding an intermediate valid record and proving the RFC 9989 walk still selects the valid record with the fewest labels. Production code is unchanged and CI is green. No remaining blocker from my review.

cristina479 (approved):

PR #1186, opened 2026-09-08
Treat DLP match counts as optional in a rule condition

### LINKED ISSUE #1185: Traceback when parsing policies with no DLP match count
### Prerequisites

### ScubaGoggles Version

ScubaGoggles v1.0.1

### Operating System

Windows 11 Enterprise 25H2

### Python Version

v3.14.7

### 🐛 Summary

When running "scubagoggles gws" the scan will fail with a traceback indicating an issue with the DLP parser noting that certain policies in our Workspace do not have match counts. Two policies in our environment are adding restraints when certain labels are applied and do not have the option for a match count to be set. Below are screenshots showing the difference from one example policy showing no option for match count, compared with one that applies a label based on a predefined data match count:

Image

Image

### Steps to reproduce

1. Open a Terminal and change directory to ScubaGoggles' Python virtual environment location, in the below example this is "C:\scuba-env"
2. Activate Python virtual environment by running "scuba-env\Scripts\activate"
3. Run "scubagoggles gws" or include any other paramters, the output below runs "scubagoggles gws --config C:\Users\username\\.scubagoggles\scubagoggles_config.yaml"

Traceback from Terminal:
```
(scuba-env) PS C:\scuba-env> scubagoggles gws --config C:\Users\username\.scubagoggles\scubagoggles_config.yaml
Traceback (most recent call last):
  File "<frozen runpy>", line 203, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "C:\scuba-env\scuba-env\Scripts\scubagoggles.exe\__main__.py", line 5, in <module>
    sys.exit(dive())
             ~~~~^^
  File "C:\scuba-env\scuba-env\Lib\site-packages\scubagoggles\main.py", line 531, in dive
    args.dispatch(args)
    ~~~~~~~~~~~~~^^^^^^
  File "C:\scuba-env\scuba-env\Lib\site-packages\scubagoggles\main.py", line 45, in gws_dispatch
    Orchestrator(args).start_automation()
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^
  File "C:\scuba-env\scuba-env\Lib\site-packages\scubagoggles\orchestrator.py", line 1005, in start_automation
    self._run_gws_providers()
    ~~~~~~~~~~~~~~~~~~~~~~~^^
  File "C:\scuba-env\scuba-env\Lib\site-packages\scubagoggles\orchestrator.py", line 232, in _run_gws_providers
    provider_dict = provider.call_gws_providers(baselines, quiet)
  File "C:\scuba-env\scuba-env\Lib\site-packages\scubagoggles\provider.py", line 1184, in call_gws_providers
    policies = policy_api.get_policies()
  File "C:\scuba-env\scuba-env\Lib\site-packages\scubagoggles\policy_api.py", line 530, in get_policies
    self._custom_parse_policies(result)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^
  File "C:\scuba-env\scuba-env\Lib\site-packages\scubagoggles\policy_api.py", line 889, in _custom_parse_policies
    parser(orgunit, section)
    ~~~~~~^^^^^^^^^^^^^^^^^^
  File "C:\scuba-env\scuba-env\Lib\site-packages\scubagoggles\parsers\dlp_rules_parser.py", line 90, in __call__
    self.dlp_rules(orgunit, section)
    ~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^
  File "C:\scuba-env\scuba-env\Lib\site-packages\scubagoggles\parsers\dlp_rules_parser.py", line 124, in dlp_rules
    detectors = self._check_condition(rule)
  File "C:\scuba-env\scuba-env\Lib\site-packages\scubagoggles\parsers\dlp_rules_parser.py", line 191, in _check_condition
    arguments_ok, detector = eval(f'cls._check_arguments{arguments}')
                             ~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<string>", line 1, in <module>
TypeError: DlpRulesParser._check_arguments() missing 1 required positional argument: 'match_counts'
```

### Expected behavior

ScubaGoggles continues, outputting a final complete report to review.

arpitjain099:
Closes #1185.

`_check_arguments` takes `match_counts` positionally with no default, and `_check_condition` calls it by evaluating the condition's own argument list. A DLP term that carries no match counts therefore arrives with two arguments instead of three, and the `TypeError` propagates all the way out of `start_automation`, so one such policy ends the entire scan rather than that single rule.

I reproduced it by handing `_check_condition` the two-argument form directly, and got the same line the report shows:

    TypeError: DlpRulesParser._check_arguments() missing 1 required positional argument: 'match_counts'

`match_counts` now defaults to `None`, and the two counts are read with `.get(..., 1)` rather than by subscript. That second part matters as much as the first: the counts are omitted individually as well as together, so a condition carrying only `minimum_match_count` would have raised `KeyError` on the next line even once the `TypeError` was gone.

On the semantics, an absent count is treated as meeting the requirement rather than failing it. A term that sets no threshold triggers on a single match, which is what the baseline asks for when it requires counts of 1, so reading the omission as a failure would report a policy as non-compliant when it is in fact stricter. That is a judgement rather than something the API documents, so if you would rather an unspecified count be treated as unknown and the rule skipped, say so and I will change it.

The new `test_dlp_rules_parser.py` covers the four shapes the condition can take (both counts, neither, and each one alone) and three that should still be rejected: a count above one, a likelihood below `LIKELY`, and a detector outside `_minimum_detectors`. Against the current code three of the seven fail with the `TypeError` above. `pytest` over `Testing/Unit/Python/parsers` and `test_policy_api.py` is 64 passed, 28 subtests passed, and pylint rates both files 10.00/10.

### DIFF
--- a/scubagoggles/Testing/Unit/Python/parsers/test_dlp_rules_parser.py
+++ b/scubagoggles/Testing/Unit/Python/parsers/test_dlp_rules_parser.py
@@ -1,0 +1,69 @@
+"""Tests for the DlpRulesParser class.
+"""
+
+import pytest
+
+from scubagoggles.parsers.dlp_rules_parser import DlpRulesParser
+
+# pylint: disable=too-few-public-methods
+
+
+def _rule(detector: str, likelihood: str, counts: str = '') -> dict:
+
+    """Builds a rule carrying a single DLP content condition, in the form
+    Google returns from the Policy API.
+    """
+
+    condition = (f'all_content.matches_dlp_detector("{detector}", '
+                 f'google.privacy.dlp.v2.Likelihood.{likelihood}{counts}')
+    condition += ')'
+
+    return {'condition': {'contentCondition': condition}}
+
+
+class TestDlpRulesParser:
+
+    """This class contains unit tests for the DlpRulesParser class.
+    """
+
+    # The test module needs to access "internal" methods.
+    # pylint: disable=protected-access
+
+    _both_counts = ', {minimum_match_count: 1, minimum_unique_match_count: 1}'
+
+    @pytest.mark.parametrize('counts',
+                             [_both_counts,
+                              '',
+                              ', {minimum_match_count: 1}',
+                              ', {minimum_unique_match_count: 1}'],
+                             ids = ['both counts',
+                                    'no counts',
+                                    'minimum only',
+                                    'unique only'])
+    def test_condition_counts_optional(self, counts):
+
+        """A condition may leave out either match count, or both.  Google
+        omits them for rules whose action has nowhere to set one, and the
+        term then triggers on a single match, which meets the baseline.
+        """
+
+        rule = _rule('CREDIT_CARD_NUMBER', 'LIKELY', counts)
+
+        assert (DlpRulesParser._check_condition(rule)
+                == {'CREDIT_CARD_NUMBER'})
+
+    @pytest.mark.parametrize(
+        'detector,likelihood,counts',
+        [('CREDIT_CARD_NUMBER', 'LIKELY',
+          ', {minimum_match_count: 5, minimum_unique_match_count: 1}'),
+         ('CREDIT_CARD_NUMBER', 'POSSIBLE', _both_counts),
+         ('EMAIL_ADDRESS', 'LIKELY', _both_counts)],
+        ids = ['count above one', 'likelihood too low', 'detector not required'])
+    def test_condition_rejected(self, detector, likelihood, counts):
+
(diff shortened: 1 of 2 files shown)


mitchelbaker-cisa:
Hi @arpitjain099, thanks for the contribution. However, I am closing this PR because one of our developers is already investigating the root issue, #1185.

PR #1189, opened 2026-09-09
Initial submission for Diff implementation in ScubaGoggles

### LINKED ISSUE #1068: Exploratory: Implement Diff in ScubaGoggles
### Prerequisites

### 💡 Summary

Investigate to understand what is required to implement the DIF in ScubaGoggles. This work is exploratory to clarify requirements, constraints, and design options.  Will build a prototype / visual. 

### Motivation and context

This exploratory issue will create enhancement issues (that will live in the backlog for now)
The goal is to determine what is the best way to create a diff tool/checker a few things to consider:

- [ ] Should this tool be independent of Goggles/Gear and be able to be ran on either?
- [ ] Should it be a command line tool only? 
- [ ] Should it create a diff report similar to SCuBA in HTML for users to be able to read?
- [ ] How does it compare results, is it strictly pass/fail or does it go into more specific things like users which have changed or gotten extra permissions since the last run?

### Implementation notes

1. A prototype of what is envisioned or a couple different ideas which can be presented to the team

### Acceptance criteria

- [x] Scope out the DIF feature that is going to be created, including its behavior, design, 
- [x] Determine if it will be part of the Goggles suite or a separate tool similar to ScubaConnect
- [x] Create follow up issues to implement the new DIF capability

### LINKED ISSUE #1188: Diff implementation of ScubaGoggles
### Prerequisites

- [x] Assignee should meet with issue author before starting work.

### 💡 Summary

This is a sub-issue of #1068 to create a diff feature. This also contains a unit test. 

### Motivation and context

This feature has been requested by the government offices.

### Implementation notes

- This is a command line tool built into goggles.
- Implementation based on the gear equivalent https://github.com/cisagov/ScubaGear/blob/ScubaGearDiffproto/docs/execution/diff.md
- It uses the parameter 'diff' instead of 'gws' to run. This kept the consistency of command structure.
- `scubagoggles diff BEFORE.json AFTER.json`
- Output into ScubaGoggles/diff-results.
- Output overrides previous output when run.
- Output report is simple. Feedback seemed to be that something easy to parse and look for relevant information was the main goal. Additional formatting / design could make it look better, but may also take away from the minimalist nature of finding relevant information quickly.
- Also comes with a unit test for diff.py.

Other:
- used 'pylint: disable=too-many-return-statements' for two functions to maintain easy readability

### Acceptance criteria

- [ ] Does the diff tool correctly compare two different reports
- [ ] Does the implementation match the gear equivalent
- [ ] Is the unit test coverage adequate

Git-Minty:
## 🗣 Description ##

This is the prototype for the diff tool for goggles requested in #1068 and created in #1188. It takes in two report jsons, compares them, and attempts to inform the user about the differences between the two. 

This also includes a unit test.

### 💭 Motivation and context

There have been requests for this feature from the government offices. 

## 🧪 Testing

I have run this on multiple reports and output is as expected for each run. Additionally, unit test was created and passes all tests.

## ✅ Pre-merge Checklist

## ✅ Post-merge Checklist


scubagoggles/Testing/Unit/Python/test_diff.py
```diff
@@ -0,0 +1,82 @@
...
+
+    def test_result_diff(self):
+        """Classify result changes."""
+        assert result_diff("Pass", "Fail") == "NewFail"
+        assert result_diff("Fail", "Pass") == "NewPass"
+        assert result_diff("Pass", "N/A") == "NewManualCheck"
+        assert result_diff("Pass", "Omitted") == "NewOmission"
+        assert result_diff("Pass", "Error") == "Errored"
+
+    def test_split_version_is_case_insensitive(self):
```
pvanderveen110, line 47: Should contain a test case where the control id doesn't match, i.e. "GWSUNMATCHEDPOLICY"

scubagoggles/diff.py
```diff
@@ -0,0 +1,490 @@
...
+        return control_id, None
+    return match.group("base"), int(match.group("version"))
+
+
+def _parse_control(
+    product: str,
+    group: dict[str, Any],
+    raw: dict[str, Any],
+) -> Control | None:
+    """Normalize one raw control record."""
```
pvanderveen110, line 84: Perhaps some documentation on the parameters might be useful here; i.e. reading "raw" I understand it's data structure but not its context/role.

scubagoggles/diff.py
```diff
@@ -0,0 +1,490 @@
...
+    if before == "N/A" and after in AUTOMATED_STATES:
+        return "NewAutomatedCheck"
+    if before in AUTOMATED_STATES and after == "N/A":
+        return "NewManualCheck"
+    if after == "Omitted" and before != "Omitted":
+        return "NewOmission"
+    if before == "Omitted" and after != "Omitted":
+        if after == "N/A":
+            return "NewManualCheck"
+        return AUTOMATED_TRANSITIONS.get(after)
```
pvanderveen110, line 150: Why do we not have cases for "No events found" in the branch { if before == "Omitted" and after != "Omitted" } ?

scubagoggles/diff.py
```diff
@@ -0,0 +1,490 @@
...
+    if after in AUTOMATED_STATES:
+        return AUTOMATED_TRANSITIONS[after]
+    if after == "N/A":
+        return "NewManualCheck"
+    if after == "Omitted":
+        return "NewOmission"
+    return "Other"
+
+
+def _standard_transition(before: str, after: str) -> str:
```
pvanderveen110: @Git-Minty shouldn't this also include states that are manual after as well? Also, I would recommend refactoring some of these methods, particularly _error_transition, _manual_transition, _standard_transition. 

I think it might be plausible to group different before/after transitions into different labels based on their implications (i.e. NewIncorrectResult, Incorrect result, ... ).

scubagoggles/diff.py
```diff
@@ -0,0 +1,490 @@
...
+        return error_result
+
+    manual_result = _manual_transition(before, after)
+    if manual_result is not None:
+        return manual_result
+
+    return _standard_transition(before, after)
+
+
+def classify_pair(before: Control | None, after: Control | None) -> str:
```
pvanderveen110: What about when both before AND after are None?

scubagoggles/diff.py
```diff
@@ -0,0 +1,490 @@
...
+    if before is None:
+        return "NewPolicy"
+    if after is None:
+        return "RemovedPolicy"
+
+    before_base, before_version = split_version(before.control_id)
+    after_base, after_version = split_version(after.control_id)
+    if before_base.lower() != after_base.lower():
+        return "Other"
+    if (
```
pvanderveen110: Perhaps some "Other" change when versions are "None"?

pvanderveen110:
@ahuynhECS I might suggest if you can, please feel free to take a look at some of my comments to and see if they make sense or if you have anything to add. Heads up, I'm finding a lot of potential value in refactoring some of the methods. I haven't gotten through reviewing the whole thing yet, but I'm commenting as I go.

scubagoggles/diff.py
pvanderveen110: The final result_diff(before.result, after.result) seems (in context) only to be executed if the controls and versions are the same (to search for deeper modifications).

scubagoggles/diff.py
```diff
@@ -0,0 +1,490 @@
...
+        "SchemaVersion": SCHEMA_VERSION,
+        "Tool": "ScubaGogglesDiff",
+        "Before": before_report.get("MetaData", {}),
+        "After": after_report.get("MetaData", {}),
+        "Summary": dict(summary),
+        "Records": records,
+    }
+
+
+def write_json(result: dict[str, Any], path: Path) -> None:
```
pvanderveen110: Left off here. @pvanderveen110

scubagoggles/Testing/Unit/Python/test_diff.py
Git-Minty: Agreed. Added a separate test to test this since is testing a different behavior.

scubagoggles/diff.py
Git-Minty: Agreed. Changed raw to control_data, and added some documentation.

scubagoggles/diff.py
Git-Minty: Good catch. I think rather than having one offs, we want something with more transition coverage. Something like this, but didn't have time to think through it completely.

`def _manual_transition(before: str, after: str) -> str | None:
    """Classify transitions involving manual or omitted states."""
    if after == "Omitted":
        return "NewOmission"

    if before == "Omitted":
        if after in MANUAL_STATES:
            return "NewManualCheck"
        return AUTOMATED_TRANSITIONS.get(after)

    if before in MANUAL_STATES and after in AUTOMATED_STATES:
        return "NewAutomatedCheck"

    if before in AUTOMATED_STATES and after in MANUAL_STATES:
        return "NewManualCheck"

    return None`

pvanderveen110:
@Git-Minty I'm Done with the review comments.

scubagoggles/diff.py
```diff
@@ -0,0 +1,500 @@
...
+            return "NewManualCheck"
+        if after == "Omitted":
+            return "NewOmission"
+        return "Other"
+    if after in AUTOMATED_STATES:
+        return AUTOMATED_TRANSITIONS[after]
+    return "Other"
+
+
+def result_diff(before: str, after: str) -> str:
```
pvanderveen110: Todo: Refactor as 

def result_diff(before: str, after: str) -> str:
    """Classify a change between two control results."""
    if before == after:
        return "Unchanged"
    return classify_diff(before, after)  #<--- one refactored method

pvanderveen110: i.e method doesn't have to be classify_diff but should contain any possible differential status (manual, automatic, error, etc.).

pvanderveen110: new idea to just implement the refactoring in results_diff

pvanderveen110 (approved):
Added refactoring, the main point of focus I tried working on was following a similar precedence and diff results as with Gear (minus the Migration changes).

ahuynhECS (changes requested):
Diffed this against ScubaGear's parallel implementation to check for behavioral and presentation parity. Comments / suggestions listed below:


1. **ScubaGoggles doesn't color rows by the after-column result at all, unlike ScubaGear, and this produces a real gap for two classifications.** ScubaGear colors each row by `ResultAfter` alone (`Get-ScubaRowColorClass`), independent of classification, so a row's color always reflects what's true *right now*. ScubaGoggles colors rows by a fixed classification -> color mapping (`_classification_css`), so the color reflects *why it changed*, never *what state it's in*.

   - **`NewAutomatedCheck` renders green even when the underlying result is a Fail.** `_classification_css` groups `NewAutomatedCheck` with `NewPass` under `"good"`. But `NewAutomatedCheck` fires for *any* `N/A -> {Pass, Fail, Warning}` transition, not just `N/A -> Pass` (confirmed: `before == "N/A"` returns `NewAutomatedCheck` before the Pass/Fail/Warning lookup runs). A newly-automated **failing** control currently shows up green. ScubaGear avoids this because it colors rows by `Result (After)`, not by classification, so a Fail is always red regardless of why the row changed.

   - **`PolicyVersionUpdate` renders neutral regardless of outcome, and that's not enough information either.** Unlike `NewAutomatedCheck`, this one isn't miscolored in one fixed wrong direction, it's colored with **no signal at all**. A version-bumped control's `ResultAfter` can be anything (Pass, Fail, Warning, N/A, Omitted), since the version check is orthogonal to the result diff. ScubaGear reflects that: a version-bumped control that's now failing renders red, one that's passing renders green. ScubaGoggles renders every `PolicyVersionUpdate` row identically neutral, so a reviewer scanning for red rows will miss a version-bumped control that's currently failing, the classification badge says "this policy changed shape" but the row gives no indication whether that's currently a problem.

Recommend switching row coloring to key off `ResultAfter` to match ScubaGear, rather than trying to hand-maintain a fixed per-classification color table, that's the fix for both of these at once, not two separate patches.

2. **`GroupNumber` and control IDs sort lexicographically, not numerically** (`_record_sort_key`). Gmail's own baseline runs groups 1 through 18, and the sample report already has `GroupNumber` values "10" through "14." As written, groups/controls 10+ will sort before single digits in both the HTML and CSV output.

3. **Timestamp parser** Just added a timestamp checker `Invoke-SCuBADiff` that warns when `AfterPath`'s `MetaData.TimestampZulu` is earlier than or equal to `BeforePath`'s.

4. **Column/filter order differs** between the two tools' summary tables despite using the same 13 classification names. Recommend both converge on a single severity-ordered layout: `Errored, NewFail` (currently broken) -> `NewWarning` (degraded) -> `NewIncorrectResult, PolicyVersionUpdate, NewOmission, Other` (needs manual review) -> `NewAutomatedCheck, NewManualCheck` (coverage/automation shape changed) -> `NewPass, NewPolicy, RemovedPolicy` (good news / administrative) -> `Unchanged` (already hidden by default in both).

5. **Other report differences** that ScubaGear's HTML report has: a color legend, Before/After source metadata cards (tool version, timestamp, report UUID), "products only in Before/After" callouts, a per-product Total column in the summary table, and a Dark Mode toggle.
 
6. **Custom filenames** Also missing relative to `Invoke-SCuBADiff`: configurable output filenames (`-OutCsvFileName`/`-OutReportFileName` equivalents; currently hardcoded `DiffReport.html`/`DiffResults.csv`).

PR #1190, opened 2026-09-09
Handle tenants where the directory API omits organizationUnits

soroush5:
## 🗣 Description ##

`Provider.get_toplevel_ou()` crashed with `KeyError: 'organizationUnits'` when the directory API returns a response without that key (e.g. `{}`). Since this runs during initialization, one such tenant kills the whole ScubaGoggles run.

The sibling method `get_ous()` already handles exactly this shape (`if 'organizationUnits' not in response: return {}`), so this just brings `get_toplevel_ou()` in line with it: fall through to the existing "Unable to determine the name of the top-level OU." warning and return `''`.

## 🧪 Testing ##

- Added the `{}` response to `GET_TOPLEVEL_OU_CASES` — failed with `KeyError` before the fix, passes after.
- Full Python unit suite (excluding `ui`, which needs `streamlit`): 429 passed.


mitchelbaker-cisa:
Hi @soroush5, thanks for the contribution. Could you file a bug report with your findings and include any testing steps you've done when running ScubaGoggles against a GWS tenant? This allows us to understand the problem in more detail before jumping into implementation. I'm closing this PR for now.

PR #1191, opened 2026-09-09
Fix UnboundLocalError in get_tenant_info when the directory API fails

soroush5:
## 🗣 Description ##

`Provider.get_tenant_info()` crashed with `UnboundLocalError: primary_domain` whenever `customers().get().execute()` raised (e.g. a 403 or an API outage). The `except` handler itself references `primary_domain`, which was only assigned inside the `try` — so instead of the intended graceful `{"ID": "", "domain": "Error Retrieving", ...}` result, the whole ScubaGoggles run died. Since this runs during initialization, one API error means zero report.

This moves the `'Error Retrieving'` default above the `try` (the same shape the sibling getters already use) and updates the existing test, which had codified the crash via `pytest.raises(UnboundLocalError)`, to assert the graceful result instead.

## 🧪 Testing ##

- Reproduced locally with a mocked failing `execute()`: `UnboundLocalError` before, graceful dict after.
- `test_get_tenant_info` (4 cases) green; full Python unit suite (excluding `ui`, which needs `streamlit`): 428 passed.


mitchelbaker-cisa:
Hi @soroush5, thanks for the contribution. Could you file a bug report with your findings and include any testing steps you've done when running ScubaGoggles against a GWS tenant? This allows us to understand the problem in more detail before jumping into implementation. I'm closing this PR for now.

PR #1196, opened 2026-09-10
Move release workflow scripts into utils

### LINKED ISSUE #1184: Remove In Line Scripts from Run_Release Workflow
### Prerequisites

- [ ] Assignee should meet with issue author before starting work.

### 💡 Summary

Recent updates in the run_release workflow has added in line scripts within the workflow file. These should be moved out into a workflow file under /utils/workflow.

### Motivation and context

We shouldn't have inline scripts in workflow files because they can't be linted or tested. 

### Implementation notes

1. Move in line scripting to separate file under /utils/workflow

### Acceptance criteria

- [ ] Move in line scripting to separate file and ensure run_release workflow runs.

sylvesterkaczmarek:
## Summary

Move the inline scripts from `run_release.yml` into `utils/workflow` so they can be linted and tested independently.

- extract previous-release lookup and release-note generation to JavaScript modules
- extract packaging and release-note display shell blocks
- keep workflow steps as concise script invocations

## Validation

- workflow YAML parses successfully
- `node --check` passes for both JavaScript files
- `bash -n` passes for both shell scripts
- extracted JavaScript logic matches the original workflow blocks exactly
- no `run: |` or `script: |` blocks remain in `run_release.yml`

Fixes #1184


mitchelbaker-cisa:
Hi @sylvesterkaczmarek, thanks for the contribution. However, since the original issue #1184 involves changes to our run_release.yml workflow, this will involve hands-on testing from our internal development team to conduct a thorough review.

PR #1198, opened 2026-09-10
Keep user configuration defaults independent between instances

### LINKED ISSUE #1197: UserConfig instances share mutable default settings
### Prerequisites

### ScubaGoggles Version

Current main, `2dda8c88fd021e782be93d317af5086f59cc508a` (1.0.1).

### Operating System

macOS; the affected code is platform-independent.

### Python Version

3.12

### Summary

`UserConfig` uses `dict(self._defaults)` when its file does not exist. This copies only the outer dictionary. The nested settings dictionary remains shared by all new instances and the class defaults.

Changing an instance's output directory, OPA directory, or credentials path also changes other instances and the defaults inherited by subsequent instances. Writing another configuration can therefore persist values that were only set on the first configuration.

### Steps to reproduce

Use two different, nonexistent configuration paths in a temporary directory:

```python
first = UserConfig(temp_dir / 'first.yaml')
second = UserConfig(temp_dir / 'second.yaml')
first.output_dir = temp_dir / 'custom'
assert second.output_dir == Path('.')  # fails
```

The same problem occurs through `opa_dir` and `credentials_file`. Three parametrized regression cases reproduce the failure without reading a real user configuration or using cloud credentials.

### Expected behavior

Each configuration should own its settings. Changes should affect only that instance and the file it explicitly writes.

sylvesterkaczmarek:
## Summary

Deep-copy default settings when creating a `UserConfig` without an existing file. Changes to output, OPA or credentials paths now stay within that instance instead of modifying other configurations and the class defaults.

Fixes #1197.

## Testing

- Added three parametrized regression cases covering existing instances, subsequent instances and independent file round-trips. All three fail before the fix.
- Full Python unit suite on Python 3.12: **582 passed, 33 subtests passed**.
- Changed-file Pylint passes with the repository's disabled-check list.

### DIFF
--- a/scubagoggles/Testing/Unit/Python/config/test_config.py
+++ b/scubagoggles/Testing/Unit/Python/config/test_config.py
@@ -118,3 +118,25 @@ class TestUserConfig:
 
         # Verify the error message contains the invalid key
         assert 'invalid_key' in str(exc_info.value)
+
+    @pytest.mark.parametrize('setting', ['output_dir', 'opa_dir', 'credentials_file'])
+    def test_new_configs_do_not_share_settings(self, tmp_path, setting, mocker):
+        """Changing one configuration must not alter other instances or defaults."""
+        mocker.patch.object(UserConfig, '_transition_config_file')
+        first = UserConfig(tmp_path / 'first.yaml')
+        second = UserConfig(tmp_path / 'second.yaml')
+        original_value = getattr(second, setting)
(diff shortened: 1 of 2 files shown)


mitchelbaker-cisa:
Closing this PR as the associated issue https://github.com/cisagov/ScubaGoggles/issues/1197 is not in our current milestone backlog and needs to be triaged before its implementation begins.

PR #1199, opened 2026-09-10
Preserve executable paths in the OPA download check

sylvesterkaczmarek:
## Summary

Pass the downloaded OPA executable as one absolute argument when running its version check. The previous whitespace split breaks paths such as `OPA Tools/opa`, and a relative basename can be resolved through PATH instead of the download directory.

The change preserves existing error logging and does not introduce shell execution.

## Testing

- Python 3.12 on Ubuntu: **591 tests and 33 subtests pass** in the complete Python unit suite.
- All **12 new regression cases pass**; eight fail with the original implementation.
- Cases cover spaces, brackets, apostrophes, relative basenames, nonzero exits and missing executables. Five cases launch real local fixtures without downloads.
- Changed-file Pylint passes.
- Validation run: https://github.com/sylvesterkaczmarek/ScubaGoggles/actions/runs/34509253098

Windows execution was not tested. The temporary validation workflow is on a separate branch and is not part of this PR.

### DIFF
--- a/scubagoggles/Testing/Unit/Python/test_getopa_paths.py
+++ b/scubagoggles/Testing/Unit/Python/test_getopa_paths.py
@@ -1,0 +1,64 @@
+"""Regression tests for OPA executable paths used by the download smoke check."""
+
+import logging
+import os
+import stat
+import sys
+from pathlib import Path
+from types import SimpleNamespace
+from unittest.mock import patch
+
+import pytest
+
+from scubagoggles import getopa
+
+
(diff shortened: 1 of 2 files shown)


mitchelbaker-cisa:
Hi @sylvesterkaczmarek, there is not an associated issue for this PR which makes it difficult to review these proposed changes. Please create an issue first and our team will determine its priority and inclusion in the project from there.

sylvesterkaczmarek:
Created #1202 for triage. Thanks.

PR #1200, opened 2026-09-10
Load custom configurations when the legacy defaults file is absent

sylvesterkaczmarek:
## Summary

Check that the legacy defaults file exists before comparing it with a supplied configuration. On a fresh installation, an existing custom YAML file currently raises `FileNotFoundError` in `Path.samefile()` because `~/.scubagoggles` does not exist.

Preserve legacy migration, current default-directory loading and missing-custom-file behavior. Separate from #1198.

## Testing

- All 586 Python tests and 33 subtests pass on Python 3.12.
- Seven new regression cases cover string, Path and environment-expanded paths, missing files, current defaults, migration and an unrelated legacy file. Three cases fail before the fix.
- Changed-file Pylint, compilation and `git diff --check` pass.
- Validation: https://github.com/sylvesterkaczmarek/ScubaGoggles/actions/runs/34521863645

Tests isolate temporary configuration files and defaults. No live Google Workspace tenant or Windows execution was tested. The temporary validation workflow is excluded from this PR.

### DIFF
--- a/scubagoggles/Testing/Unit/Python/test_config_fresh_install.py
+++ b/scubagoggles/Testing/Unit/Python/test_config_fresh_install.py
@@ -1,0 +1,94 @@
+"""Configuration loading when the legacy defaults file is absent."""
+
+from pathlib import Path
+
+import pytest
+import yaml
+
+from scubagoggles.config import UserConfig
+
+
+@pytest.fixture
+def isolated_defaults(tmp_path, monkeypatch):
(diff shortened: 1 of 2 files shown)


mitchelbaker-cisa:
Hi @sylvesterkaczmarek, there is not an associated issue for this PR which makes it difficult to review these proposed changes. Please create an issue first and our team will determine its priority and inclusion in the project from there.

sylvesterkaczmarek:
Created #1201 for triage. Thanks.

PR #1206, opened 2026-09-13
Fix GWS init crashes: missing organizationUnits key and UnboundLocalError in get_tenant_info

### LINKED ISSUE #1204: GWS init crashes: missing organizationUnits key and UnboundLocalError in get_tenant_info
### Prerequisites
- [X] This issue has an informative and human-readable title.

### ScubaGoggles Version
1.0.1 (main @ dd0e34a)

### Operating System
macOS 26.6.1

### Python Version
v3.12 (test env)

### 🐛 Summary
Two crashes during initialization kill the whole ScubaGoggles run when the Google directory API returns an unexpected shape:

1. `Provider.get_toplevel_ou()` raises `KeyError: 'organizationUnits'` when the response has no such key (e.g. `{}`). The sibling `get_ous()` already guards this (`if 'organizationUnits' not in response: return {}`).
2. `Provider.get_tenant_info()` raises `UnboundLocalError: primary_domain` when `customers().get().execute()` raises (e.g. a 403): the `except` handler references `primary_domain`, which is only assigned inside the `try`, so the intended graceful `{"ID": "", "domain": "Error Retrieving", ...}` result never happens.

### Steps to reproduce
1. Mock the directory API `customers().get().execute()` to raise (for #2), or return `{}` for the OU call (for #1).
2. Run the GWS provider init path.
3. Full run dies with `UnboundLocalError` / `KeyError` instead of a warning + graceful result.

I found these through code inspection plus mocked-API repros and the unit suite (429 passed excluding `ui`, which needs `streamlit`). I have not run ScubaGoggles against a live GWS tenant, so I can't say how often real tenants hit these shapes — happy to adjust the report if these responses can't occur in practice.

### Expected behavior
A warning plus a graceful fallback (empty OU name / "Error Retrieving" tenant info), letting the rest of the report generate.

Anai-Guo:
## Summary

Fixes #1204. Two crashes during GWS provider initialization take down the whole ScubaGoggles run when the Google directory API returns an unexpected shape:

1. **`Provider.get_toplevel_ou()` — `KeyError: 'organizationUnits'`.** When the OU list response has no `organizationUnits` key (e.g. `{}`), the loop `for ou in response['organizationUnits']` raises `KeyError`. The sibling `get_ous()` already guards this case (`if 'organizationUnits' not in response: return {}`).
2. **`Provider.get_tenant_info()` — `UnboundLocalError: primary_domain`.** The `except` handler returns `{'ID': '', 'domain': primary_domain, ...}`, but `primary_domain` is only assigned *inside* the `try`. When `customers().get().execute()` raises before that line (e.g. a 403), the name is unbound and the intended graceful `{"ID": "", "domain": "Error Retrieving", ...}` result never happens.

## Fix

- `get_toplevel_ou()`: guard the missing `organizationUnits` key and return `''` with the existing "Unable to determine the name of the top-level OU." warning, mirroring the defensive style already used in `get_ous()`.
- `get_tenant_info()`: initialize `primary_domain = 'Error Retrieving'` alongside `tenant_id = ''` before the `try`, so the `except` handler always has a bound value (and the in-`try` assignment is no longer needed).

## Tests

- `admin_ou_cases.py`: added a `GET_TOPLEVEL_OU_CASES` entry for a response with the `organizationUnits` key entirely absent (`{}`).
- `test_provider.py`: `test_get_tenant_info` previously asserted `pytest.raises(UnboundLocalError)` on the API-failure cases — it now asserts the graceful result equals the `expected` dict already declared in those cases, matching the documented behavior.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


FollyBeachGurl:
Closing this PR as the associated issue is not in our current milestone backlog and needs to be triaged before its implementation begins.

